PR #7, opened 2018-12-06
Pr branch： lavc/qsvenc: replace assert with error return

fulinjie:
lavc/qsvenc: replace assert with error return

    bs->FrameType is not set in MSDK in some cases (mjpeg encode for example),
    and assert on a value coming from an external library is not proper.
    
    Add default type check for bs->FrameType, and return invalid data error in function
    ff_qsv_encode to avoid using uninitialized value.
    
    Fix #7593.


fulinjie:
patch link: https://patchwork.ffmpeg.org/patch/11326/

uartie:
merged upstream

PR #17, opened 2019-01-18
lavc/vaapi_encode_mpeg2: fix the desired packed header

fulinjie:
Only VA_ENC_PACKED_HEADER_RAW_DATA is supported as the packed header
for Mpeg2 profile in driver:

    else if (IsMpeg2Profile(profile))
    {
        attrib.value = VA_ENC_PACKED_HEADER_RAW_DATA;
    }

Modify to set the supported packed header and avoid the warning.

### DIFF
--- a/libavcodec/vaapi_encode_mpeg2.c
+++ b/libavcodec/vaapi_encode_mpeg2.c
@@ -613,8 +613,7 @@ static av_cold int vaapi_encode_mpeg2_init(AVCodecContext *avctx)
         return AVERROR(EINVAL);
     }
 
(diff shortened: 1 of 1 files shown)


fulinjie:
will trigger a regression in i965 due to the difference in driver:
i965:
allows set packed header from app level, and support:
0xb = VA_ENC_PACKED_HEADER_SEQUENCE | VA_ENC_PACKED_HEADER_PICTURE | VA_ENC_PACKED_HEADER_MISC

iHD:
// Since sequence header and picture header will be packed in codechal, we don't support packed header from application
only allows VAEncPackedHeaderRawData

colse.

PR #25, opened 2019-03-15
Pr qsvvpp  (transpose support and continuously memory allocate)

fulinjie:
patch set for qsvvpp transpose.(mirror and rotation)


fulinjie:
For transpose, MSDK has some limitation: 
- Only NV12 as in/out supported by mirroring now
- mirroring does not support crop X and Y

fulinjie:
https://github.com/intel-media-ci/ffmpeg/pull/54

PR #30, opened 2019-03-25
libavcodec/vp9: fix ref-frame size judging method

ConsoleYan:
There is no need all reference frame demension is valid in libvpx.

So this change contains three part:
1. Change each judgement's loglevel from "ERROR" to "WARNING"
2. Make sure at least one of frames that this frame references has valid dimension.
3. All judgements fail would report "ERROR".

Signed-off-by: Yan Cen <cenx.yan@intel.com>


uartie:
Has this been submitted to upstream ML?

uartie:
all checks passed. closing

PR #34, opened 2019-04-10
lavc/qsvenc: Fix the memory leak for enc_ctrl.Payload

fulinjie:
frame->enc_ctrl.Payload is malloced in get_free_frame, directly memset
the whole structure of enc_ctrl to zero will cause the memory leak for
enc_ctrl.Payload.

Fix the memory leak issue and reset other members in mfxEncodeCtrl.


fulinjie:
merged at:
http://git.ffmpeg.org/gitweb/ffmpeg.git/commit/8f6e65183354d1d402ae80c71cba2759fe152018
closed.

PR #35, opened 2019-04-10
lavu/hwcontext_qsv: Fix the realign check for hwupload

fulinjie:
Fix the aligned check in hwupload, input surface should be 16 aligned
too.

Fix #7830.


fulinjie:
merged, close.

PR #36, opened 2019-04-22
lavc/vaapi_encode_h264: add support for maxframesize

fulinjie:
Add -max_frame_size (bytes) to indicate the allowed max frame size.

Currently only AVC encoder can support this settings in multiple pass case.
If the frame size exceeds, the encoder will do more pak passes to adjust the
QP value to control the frame size.

Set Default num_passes to 4 (1~4), set delta_qp[4] = {1, 1, 1, 1}, use
new_qp for encoder if frame size exceeds the limitation:
    new_qp = base_qp + delta_qp[0] + delta_qp[1] + ...

### DIFF
--- a/libavcodec/vaapi_encode.c
+++ b/libavcodec/vaapi_encode.c
@@ -235,7 +235,15 @@ static int vaapi_encode_issue(AVCodecContext *avctx,
         if (err < 0)
             goto fail;
     }
-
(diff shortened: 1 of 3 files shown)


fulinjie:
Refined a new version.
Re-base and split mfs params buffer according to Mark T's modification;
Add attribute query for max frame size.
https://github.com/intel-media-ci/ffmpeg/pull/53

PR #37, opened 2019-04-25
lavc/qsvdec: add query function and provide error message

lizhong1008:
It is helpful to know why some clips decoding failed.
Ticket#7330 is a good example, with this patch it is easily to
know bitstream codec level is out of support range.


lizhong1008:
Patches merged. So close it.

PR #40, opened 2019-04-28
lavc/qsvenc: disable CO3 for mpeg2

lizhong1008:
Currenntly there is not any function of CO3 appled to mpeg2,
and enabling for mpeg2 it will cause regression with some old
libmfx libaries (see tiket #7839).

Also add runtime version check for CO3.


uartie:
all checks passed. closing

PR #42, opened 2019-05-08
lavc/vaapi_encode_h265: add support for low power mode for HEVC encode

fulinjie:
Enable hevc_vaapi vdenc:
 - media-driver require CTU size must be set as 64x64 at low power mode
 - cu_qp_delta_enabled_flag should be enabled for low power mode
 - low delay B to replace P frame
 - same ref_pic_list0/ref_pic_list1

Signed-off-by: Fu Linjie <linjie.fu@intel.com>


fulinjie:
issue for ICQ mode support (ICQ will be set in ffmpeg vaapi by default):
https://github.com/intel/media-driver/issues/618

-qp should be set to test CQP mode.

fulinjie:
ICQ mode is not supported in low power mode, disabled on ffmpeg-level.

uartie:
all checks passed. closing

fulinjie:
also issue for low delay B query support:https://github.com/intel/media-driver/issues/721
and CTU size requirement (internal in jira)

PR #43, opened 2019-05-17
Pr qsv 422/444

fulinjie:
add support for 422/444 qsv decode.
Currently, 8 bit 422/444 is supported.


fulinjie:
- fix the compatible issue for early libva version
- support y210/y410

uartie:
Is this the same as #57?

fulinjie:
Yes it's similar, will close this.

PR #44, opened 2019-05-29
lavc/vaapi_encode: grow packet if vaMapBuffer returns multiple buffers

fulinjie:
It seems that VA_CODED_BUF_STATUS_SINGLE_NALU allows driver to map
buffer for each slice.

Currently, assigning new buffer for pkt when multiple buffer returns
from vaMapBuffer will cover the previous encoded pkt data and lead
to encode issues.

Using av_grow_packet to expand pkt if several buffers are returned.


uartie:
all checks passed. closing

PR #45, opened 2019-05-31
Qsv vpp improvement

lizhong1008:


lizhong1008:
Patch set upstreamed.

PR #46, opened 2019-06-04
lavf/qsvvpp:allocate continuous memory

fulinjie:
Mediasdk calls CMRT to copy from video to system memory and requires
memory to be continuously allocated across Y and UV.

Add a new path to allocate continuous memory when using system out.
Use av_frame_get_buffer to arrange data according to pixfmt, and remove
the introduced palne_padding.


fulinjie:
update:
https://github.com/intel-media-ci/ffmpeg/pull/54

PR #47, opened 2019-06-07
vaapi_encode_mjpeg: fix bad component id bug

uartie:
The compound literals assigned to "components"
only exist within the scope of the if/else
block (thanks Mark Thompson for the better
explanation).

Thus, after this if/else block, "components"
ends up pointing to an arbitrary/undefined
array.  With some compilers and depending on
optimization settings, these arbitrary values
may end up being the same value (i.e. 0 with
GNU GCC 9.x).  Unfortunately, the GNU GCC
compiler, at least, never prints any warnings
about this.

This patch fixes this issue by assigning the
constant arrays to local variables at function
scope and then pointing "components" to those
as necessary.

Fixes #7915

Signed-off-by: U. Artie Eoff <ullysses.a.eoff@intel.com>


uartie:
all checks pass and patch submitted upstream.  closing.

PR #66, opened 2019-08-07
lavc/qsvdec: cope with unknown profile for HEVC

fulinjie:
Currently, unknown profile will cause the decode failure directly in
qsv.

Add a more compatible path to cope with the unknown profile. If
AV_HWACCEL_FLAG_ALLOW_PROFILE_MISMATCH flag is set, decoder will try
default mfx_profile for unknown profile and bump a warning message.

cmdline:
    ffmpeg ... -hwaccel_flags allow_profile_mismatch ...

Also modify in qsv_profile_map[] to make the code look more neat.

### DIFF
--- a/libavcodec/qsv.c
+++ b/libavcodec/qsv.c
@@ -79,7 +79,8 @@ static const struct {
     MAP(MPEG2VIDEO,  MPEG2_HIGH,      MPEG2_HIGH   ),
 
     MAP(H264,        H264_BASELINE,   AVC_BASELINE ),
-    MAP(H264,        H264_CONSTRAINED_BASELINE, AVC_BASELINE),
+    MAP(H264,        H264_CONSTRAINED_BASELINE,
+                                      AVC_BASELINE ),
 #if QSV_VERSION_ATLEAST(1, 3)
     MAP(H264,        H264_EXTENDED,   AVC_EXTENDED ),
 #endif
@@ -90,11 +91,13 @@ static const struct {
 #if QSV_VERSION_ATLEAST(1, 8)
     MAP(HEVC,        HEVC_MAIN,       HEVC_MAIN    ),
     MAP(HEVC,        HEVC_MAIN_10,    HEVC_MAIN10  ),
-    MAP(HEVC,        HEVC_MAIN_STILL_PICTURE,    HEVC_MAINSP ),
+    MAP(HEVC,        HEVC_MAIN_STILL_PICTURE,
+                                      HEVC_MAINSP  ),
 #endif
 #if QSV_VERSION_ATLEAST(1, 16)
     MAP(HEVC,        HEVC_REXT,       HEVC_REXT    ),
 #endif
+    MAP(HEVC,        UNKNOWN,         HEVC_MAIN    ),
 
     MAP(VC1,         VC1_SIMPLE,      VC1_SIMPLE   ),
     MAP(VC1,         VC1_MAIN,        VC1_MAIN     ),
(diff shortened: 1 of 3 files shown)


libavcodec/qsv.c
```diff
@@ -79,7 +79,8 @@ static const struct {
     MAP(MPEG2VIDEO,  MPEG2_HIGH,      MPEG2_HIGH   ),
 
     MAP(H264,        H264_BASELINE,   AVC_BASELINE ),
-    MAP(H264,        H264_CONSTRAINED_BASELINE, AVC_BASELINE),
+    MAP(H264,        H264_CONSTRAINED_BASELINE,
+                                      AVC_BASELINE ),
```
lizhong1008, line 83: please don't miss such thing in one patch.  One separated patch will be better.

libavcodec/qsv.c
```diff
@@ -103,17 +106,33 @@ static const struct {
...
+        if (avctx->hwaccel_flags & AV_HWACCEL_FLAG_ALLOW_PROFILE_MISMATCH &&
+            qsv_profile_map[i].codec_profile == FF_PROFILE_UNKNOWN ) {
+
+            av_log(avctx, AV_LOG_VERBOSE, "Codec %s profile %d not "
+                "supported for hardware decode.\n", codec_desc->name, profile);
+            av_log(avctx, AV_LOG_WARNING, "Using possibly-incompatible "
+                "mfx profile %d instead.\n", qsv_profile_map[i].mfx_profile);
+
+            return qsv_profile_map[i].mfx_profile;
+        }
```
lizhong1008, line 135: If this is really need to be fix, would better to fix in MSDK instead of ffmpeg-qsv. You will always set a main profile for unknown profile case but it may be main10 profile actually.

fulinjie: > If this is really need to be fix, would better to fix in MSDK instead of ffmpeg-qsv. 
Since a specific flag AV_HWACCEL_FLAG_ALLOW_PROFILE_MISMATCH is set by user,  I thought it's acceptable to use a default value for hwaccel decode in this compatible mode. 
And there is a same path in vaapi, so I thought it will make sense.
https://github.com/FFmpeg/FFmpeg/blob/6c67c8ca9ae603b91e65274f93099ccb07bb33db/libavcodec/vaapi_decode.c#L479


>You will always set a main profile for unknown profile case but it may be main10 profile actually.

Yes, i've thought about this. And desc->comp[0].depth could be used to determine the possible profile (main or main10), and other desc information could also be parsed to guess the potential profile for HEVC_REXT, and it may be too complex for current stage.

Actually, after confirming with QA teams, I eliminate the depth relevant code because there is no profile 0 cases in HEVC 10 bit.

libavcodec/qsv.c
fulinjie: Sure, thanks.

libavcodec/qsv.c
fulinjie: It could be handled in MSDK in UMC_HEVC_DECODER::MFX_Utility::CheckVideoParam_H265():
https://github.com/Intel-Media-SDK/MediaSDK/blob/5b23796ab993771718fec3b997c3f4a5c98c6c4c/_studio/shared/umc/codec/h265_dec/src/umc_h265_mfx_utils.cpp#L987
But another question may be how can user control/know whether QSV uses a compatible profile in ffmpeg level?
Silently changing the profile in MSDK may not be good, and we need some feedback from MSDK.

uartie:

fulinjie:
MSDK's parser should be able to cope with the unknown profile, closed

PR #67, opened 2019-08-07
lavf/scale_qsv: change alignment to be 16 bytes

lizhong1008:
32 bytes alignment is not needed and increases the failure possibilty of
SFC (low power scaling mode)


uartie:

lizhong1008:
close it since upstreamed.

PR #71, opened 2019-08-16
lavfi/qsvvpp: disable pass through mode if format changed

lizhong1008:
Fix tiket#8065


lizhong1008:
Patch set upstreamed.

PR #82, opened 2019-09-04
Qsv add system memory support

lizhong1008:


fulinjie:
Errors reported in smoke test:
"Error added: AssertionError: It appears that the QSV plugin did not load"

Since the result seems to be good, is there any update/modification needed in the smoke-test/vaapi_fits?

lizhong1008:
@fulinjie , good check. 

It just due to the “Initialize MFX session” message removed in https://github.com/intel-media-ci/ffmpeg/pull/82/files#diff-fd2ca58a1bd9374f47ccf4fe06e00029L1162

See : https://github.com/zhoujd/vaapi-fits/blob/master/test/ffmpeg-qsv/decode/decoder.py#L70


I have updated the patch and add the message back.

lizhong1008:
All checks have passed, so close it.

PR #93, opened 2019-09-16
lavu/pixfmt: add new pixel format ayuv/y210/y410

fulinjie:
Previously, media driver provided planar format(like 420 8 bit),
but for HEVC Range Extension (422/444 8/10 bit), the decoded image
is produced in packed format because Windows expects it.

Add some packed pixel formats for hardware decode support in VAAPI
and QSV:

4:2:2 10 bit: Y210
4:4:4  8 bit: AYUV
4:4:4 10 bit: Y410

Add query support in fate.


fulinjie:
closed, all checks passed.

PR #94, opened 2019-09-18
lavf/vf_deinterlace_vaapi: flush queued frame for field in Deinterlac…

fulinjie:
…ingBob

For DeinterlacingBob mode with rate=field, the frame number of output
should equal 2x input total since only intra deinterlace is used.

Currently, the first frame is queued in filed mode and won't be
outputed. Due to the async without flush, frame number of output is
[expected_number - 2].

Specifically, if the input only has 1 frame, the output will be empty.

Add deint_vaapi_request_frame for deinterlace_vaapi, send NULL frame
to flush the queued frame.

For 1 frame input in Bob mode with rate=field,
before patch: 0 frame;
after  patch: 2 frames;

ffmpeg -hwaccel vaapi -vaapi_device /dev/dri/renderD128 -v verbose -c:v
h264 -i ./input.h264 -vf 'format=nv12|vaapi,hwupload,
        deinterlace_vaapi=mode=bob:rate=field,hwdownload,format=nv12'
-pix_fmt yuv420p -f rawvideo -vsync passthrough -y dump.yuv

### DIFF
--- a/libavfilter/vf_deinterlace_vaapi.c
+++ b/libavfilter/vf_deinterlace_vaapi.c
@@ -48,6 +48,9 @@ typedef struct DeintVAAPIContext {
     int                queue_count;
     AVFrame           *frame_queue[MAX_REFERENCES];
     int                extra_delay_for_timestamps;
+
+    int                eof;
+    int                prev_pts;
 } DeintVAAPIContext;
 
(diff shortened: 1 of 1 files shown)


fulinjie:
closed, all checks passed.

PR #97, opened 2019-09-20
Dnnforpatch

guoyejun:


uartie:
The failures are not caused by this patch.  They are due to upstream regression in MSDK https://github.com/Intel-Media-SDK/MediaSDK/pull/1653

PR #103, opened 2019-09-23
Pr  vaapi hevc 444 encode

fulinjie:
Add encode support for HEVC Main444.

Encode cmdline is good but the output has the wrong resolutions with garbage.

This is blocked by the same issue in media driver reported by gst-vaapi.

This is a temporary solution since a modification for the RT format in libva is in need as well.


fulinjie:
All check passed, close this temporary solution.

fulinjie:
number of component would be required to compare with the supported profile from driver:
ctx->codec->profiles[i] = {av_profile = 4, depth = 8, nb_components = 4, log2_chroma_w = 0, log2_chroma_h = 0, va_profile = VAProfileHEVCMain444

Hence, the nb_components should be 4.

AYUV matches, however, 0yuv is declared to have 3 components, that's a big conflict.

ffmpeg -hwaccel vaapi -vaapi_device /dev/dri/renderD128 -v verbose -f rawvideo -video_size 1920x1080 -pix_fmt ayuv -i ./ayuv.yuv -an -vf hwupload=extra_hw_frames=64 -c:v hevc_vaapi -profile rext -level 62 -b:v 4M -low_power 1 -vframes 1 -y ./vaapi.h265

PR #115, opened 2019-10-15
lavc/qsvenc: Fix bitrate_limit to allow AVC encode in limited bitrate

fulinjie:
MFXVideoENCODE_Query calls CheckVideoParamQueryLike in MSDK and
will determine whether to set param.mfx.TargetKbps to the allowed
minTargetKbps according to the bitrate_limit in extco2 buffer.

Thus q->param.ExtParam must be set before MFXVideoENCODE_Query in case
minTargetKbps is written to TargetKbps by default.

1080P AVC encoding with option "-bitrate_limit 0 -b:v 100k":
Before patch:
            902 kbps
After patch:
            156 kbps

Reported-by: Mann, J

### DIFF
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -1052,25 +1052,6 @@ int ff_qsv_enc_init(AVCodecContext *avctx, QSVEncContext *q)
     if (ret < 0)
         return ret;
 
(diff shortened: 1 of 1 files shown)


fulinjie:
All check passed, closed.
Patch:
https://patchwork.ffmpeg.org/patch/15756/

PR #120, opened 2019-10-31
lavc/qsvenc_hevc: improve the default setting

fulinjie:
Match the default setting in x265 to use 4 B-frames.

Fix #8328.


fulinjie:
All check passed, closed.

PR #121, opened 2019-11-05
lavc/vaapi_decode: set initial pool size according to dpb_size

fulinjie:
Set initial_pool_size dynamically according to dpb_size.

Compared to a fixed pool size for H.264 and HEVC, the usage of
GPU memory could be improved.

### DIFF
--- a/libavcodec/vaapi_decode.c
+++ b/libavcodec/vaapi_decode.c
@@ -408,7 +408,8 @@ static const struct {
 static int vaapi_decode_make_config(AVCodecContext *avctx,
                                     AVBufferRef *device_ref,
                                     VAConfigID *va_config,
-                                    AVBufferRef *frames_ref)
+                                    AVBufferRef *frames_ref,
+                                    int dpb_size)
 {
     AVVAAPIHWConfig       *hwconfig    = NULL;
     AVHWFramesConstraints *constraints = NULL;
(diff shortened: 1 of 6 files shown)


libavcodec/vaapi_decode.c
```diff
@@ -584,7 +570,8 @@ static int vaapi_decode_make_config(AVCodecContext *avctx,
 }
 
 int ff_vaapi_common_frame_params(AVCodecContext *avctx,
```
xuguangxin: How about add new function, ff_vaapi_common_frame_params_with_pdb_size. then ff_vaapi_common_frame_params will call ff_vaapi_common_frame_params_with_pdb_size with size dpb_size = 2.
In this way. you do not need change any code in mpeg2, vc1.... etc.

libavcodec/vaapi_decode.c
```diff
@@ -666,7 +653,8 @@ int ff_vaapi_decode_init(AVCodecContext *avctx)
     ctx->hwctx  = ctx->device->hwctx;
 
     err = vaapi_decode_make_config(avctx, ctx->frames->device_ref,
-                                   &ctx->va_config, avctx->hw_frames_ctx);
+                                   &ctx->va_config, avctx->hw_frames_ctx,
```
xuguangxin, line 664: this changed the original behavious.  It will give 16 to mpeg2. but the previous code only gives 2.

xuguangxin (changes requested):

xuguangxin (approved):
great! thanks Lingjie.

fulinjie:
All check passed, closed.

PR #128, opened 2019-11-13
lavc/vaapi_encode: reconfigure sequence parameter in case it changes

fulinjie:
Aspect ratio change should be preserved in the output result of
transcoding. A better choice is to reconfigure when parameter changing
is detected, but a fix to reconfigure on IDR frame is simple and effective
with no performance drop.

Fix #6276 for VAAPI.


fulinjie:
closed.
https://patchwork.ffmpeg.org/patch/16245/

PR #135, opened 2019-12-02
lavu/hwcontext_vaapi: Check VA_SURFACE_ATTRIB_SETTABLE for attributes

fulinjie:
VASurfaceAttribPixelFormat is used to guarantee a proper fourcc or
media_format of the created surface.

If VA_SURFACE_ATTRIB_SETTABLE is not set, these attributes may be
ignored.

Add a check for attributes.flags when creating surface to make sure
these attribuates are truly accessible.

See:
https://github.com/intel/media-driver/blob/master/media_driver/linux/
common/ddi/media_libva.cpp#L2149


fulinjie:
closed.
Need to be applied with:
https://patchwork.ffmpeg.org/patch/16332/

PR #141, opened 2019-12-10
avcodec/cbs_av1: avoid reading trailing bits when obu type is OBU_TIL…

feiwan1:
…E_LIST

Signed-off-by: Wangfei <fei.w.wang@intel.com>


feiwan1:
merged. close.

PR #143, opened 2019-12-11
Pr hevc sw decode upstream

fulinjie:


libavcodec/x86/hevc_add_res.asm
```diff
@@ -28,23 +28,23 @@ cextern pw_1023
 
 ; the add_res macros and functions were largely inspired by h264_idct.asm from the x264 project
 %macro ADD_RES_MMX_4_8 0
+    pxor              m4, m4
```
xuguangxin: move this to out of loop.

fulinjie:
all check passed, closed.

PR #144, opened 2019-12-13
avcodec/cbs_av1: avoid reset some members of loop filter

feiwan1:
According AV1 spec 6.8.10, loop_filter_ref_deltas & loop_filter_mode_deltas
should be keep same as its previous value if it is not presented in current
syntax.


zhoujd:
can we close this PR ? you can submit a new PR if you have update

feiwan1:
patch already submitted to ffmpeg, so close this PR.

PR #146, opened 2019-12-16
Pr qsv msdk

fulinjie:


dvrogozh (changes requested):
Please, set qmax/qmin defaults for qsv_hevc.c encoder in the same way we set for qsv_avc.c:
https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsvenc_h264.c#L169

I.e.:
```
    { "qmin",      "-1"    },
    { "qmax",      "-1"    },
```
Otherwise, we get ffmpeg common defaults which are qmin=2 and qmax=31 which spoil all the fun we can get from extbrc.

PR #149, opened 2019-12-24
lavc/pthread_frame: Update user context in ff_frame_thread_free

fulinjie:
Resolution/format changes lead to reinitialization of hardware
accelerations(vaapi/dxva2/nvdec) with new hwaccel_priv_data in
the worker-thread. And hwaccel_priv_data in user context won't
be updated until the resolution changing frame is output.

A termination with "-vframes" just after the reinit will lead to:
    1. memory leak in worker-thread.
    2. double free in user-thread while calling avcodec_close().

To reproduce:

ffmpeg -hwaccel vaapi(dxva2/nvdec) -v verbose -i
    fate-suite/h264/reinit-large_420_8-to-small_420_8.h264 -pix_fmt nv12
    -f rawvideo -vsync passthrough -vframes 45 -y out.yuv

### DIFF
--- a/libavcodec/pthread_frame.c
+++ b/libavcodec/pthread_frame.c
@@ -657,6 +657,13 @@ void ff_frame_thread_free(AVCodecContext *avctx, int thread_count)
 
     park_frame_worker_threads(fctx, thread_count);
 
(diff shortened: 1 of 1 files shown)


fulinjie:
all check passed, closed.

PR #156, opened 2020-01-04
lavu/hwcontext_vaapi: remove redundant check in vaapi_map_to_memory

fulinjie:
vaapi_get_image_format() will be called in vaapi_map_frame(), so it's
redundant here. Also, NULL for VAImageFormat **image_format could be
a bit improper since the function is supposed to be used to get the
image format.

Remove vaapi_get_image_format() in vaapi_map_to_memory().


fulinjie:
closed

PR #165, opened 2020-01-21
lavfi/vf_scale_qsv: add extra_hw_frames support

fulinjie:
While using scale_qsv together with lookahead, extra_hw_frames is
required to allocate more memory.

Fix #8379.


fulinjie:
It's unnecessary, extra_hw_frames could be passed by:
 -vf scale_qsv=w=1280:h=720:extra_hw_frames=64

PR #167, opened 2020-02-12
lavc/qsvenc: add support for external bitrate control for HEVC

fulinjie:
Enables option for hevc_qsv encoder:
    -extbrc

Encocoding quality improvement could be observed.

### DIFF
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -681,11 +681,8 @@ FF_ENABLE_DEPRECATION_WARNINGS
 
         q->extparam_internal[q->nb_extparam_internal++] = (mfxExtBuffer *)&q->extco;
 
-        if (avctx->codec_id == AV_CODEC_ID_H264) {
 #if QSV_HAVE_CO2
-            q->extco2.Header.BufferId     = MFX_EXTBUFF_CODING_OPTION2;
-            q->extco2.Header.BufferSz     = sizeof(q->extco2);
-
+        if (avctx->codec_id == AV_CODEC_ID_H264) {
             if (q->int_ref_type >= 0)
                 q->extco2.IntRefType = q->int_ref_type;
             if (q->int_ref_cycle_size >= 0)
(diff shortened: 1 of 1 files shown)


libavcodec/qsvenc.c
```diff
@@ -681,11 +681,10 @@ FF_ENABLE_DEPRECATION_WARNINGS
 
         q->extparam_internal[q->nb_extparam_internal++] = (mfxExtBuffer *)&q->extco;
 
-        if (avctx->codec_id == AV_CODEC_ID_H264) {
 #if QSV_HAVE_CO2
-            q->extco2.Header.BufferId     = MFX_EXTBUFF_CODING_OPTION2;
-            q->extco2.Header.BufferSz     = sizeof(q->extco2);
-
+        q->extco2.Header.BufferId     = MFX_EXTBUFF_CODING_OPTION2;
```
dvrogozh: Do I read the code correctly that you have allowed co2 for all encoders including VP9 and Jpeg thought you don't program anything in it? I don't think it will harm anything, but someone in community might consider this to be a problem.

fulinjie: Thanks for pointing this out, 
For VP9, co/co2 will be set only if avctx->codec_id != AV_CODEC_ID_VP9:
https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsvenc.c#L649

For Jpeg, ffmpeg calls init_video_param_jpeg() instead;
https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsvenc.c#L418

So these codec won't be a problem, but mpeg2_qsv does have this concern if co2 is not supported for mpeg2.

BTW, co3 has the same issue.(allowed co3 for encoder like H264, but don't program anything)
https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsvenc.c#L764

fulinjie:
https://unofficial.patchwork-ffmpeg.org/project/FFmpeg/patch/1581598044-7703-1-git-send-email-linjie.fu@intel.com/

closed.

PR #178, opened 2020-02-28
lavfi/vf_vpp_qsv: fix the infinite loop while framerate lower than input

fulinjie:
There are frame droppings in frc while converting into a lower framerate,
and MSDK returns ERROR_MORE_DATA, which could be ignored.

Reported-by: Gupta, Pallavi <pallavi.gupta@intel.com>


fulinjie:
Passed and closed.

PR #181, opened 2020-03-11
lavc/vp9: fix reference frame dimensions check

fulinjie:
With the description in frame size with refs semantics (SPEC 7.2.5),
it is a requirement of bitstream conformance that for at least one
reference frame has the valid dimensions.

Modify the check to make sure the decoder works well in the condition
that not all references frames have valid dimensions.


fulinjie:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/1584456838-7151-1-git-send-email-linjie.fu@intel.com/

PR #189, opened 2020-03-31
lavc/vaapi_encode: add FF_CODEC_CAP_INIT_CLEANUP caps for encoders

fulinjie:
ff_vaapi_encode_close() is not enough to free the resources like cbs
if initialization failure happens after codec->configure (except for
vp8/vp9).

We need to call avctx->codec->close() to deallocate, otherwise memory
leak happens.

Add FF_CODEC_CAP_INIT_CLEANUP for vaapi encoders and deallocate the
resources at free_and_end inside avcodec_open2().


fulinjie:
upstreaming, closed.

PR #204, opened 2020-05-25
dnn: add openvino as one of dnn backend

guoyejun:
OpenVINO is a Deep Learning Deployment Toolkit at
https://github.com/openvinotoolkit/openvino, it supports CPU, GPU
and heterogeneous plugins to accelerate deep learning inferencing,
and also has C API.


uartie:
all checks pass. closing

PR #211, opened 2020-06-03
Dnn add native trigonometric funcs

Semmer2:
add missing return in tan support


guoyejun:
looks good to me, thanks.

PR #216, opened 2020-06-16
Dnn add native inverse tri funcs

Semmer2:
Local MD5 test and make fate passed. Thank you for reviewing.


guoyejun:
looks good to me.

PR #220, opened 2020-06-22
Pr vaapi 422 encode recon

fulinjie:


fulinjie:
All check passed, closed.

PR #234, opened 2020-07-13
dnn/native: add native support for avg_pool

Semmer2:
passed all local MD5 test


Semmer2:
create new PR with new code

PR #257, opened 2020-08-18
qsv: add ${includedir}/mfx to the search path for old version of libmfx

xhaihao:
${inludedir}/mfx has been added to Cflags in libmfx.pc in new versions
of libmfx. We may add ${includedir}/mfx to the search path for old
version of libmfx so that we may inlcude mxfxxx.h instead of
mfx/mfxxxx.h in FFmpeg

If your libmfx comes without pkg-config support, this patch requires a
small change to your environment setting (e.g. adding
/opt/intel/mediasdk/include/mfx instead of /opt/intel/mediasdk/include
to CFLAGS) so that the build can find the headers

### DIFF
--- a/configure
+++ b/configure
@@ -1458,6 +1458,14 @@ check_pkg_config(){
         eval add_cflags \$${name}_cflags
 }
 
+append_subdir_to_pkg_includedir(){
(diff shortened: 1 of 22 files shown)


xhaihao:
check passed.

PR #260, opened 2020-08-18
Verify pr procedure

MingyuY:


libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -27,6 +27,64 @@
 #include "libavutil/avassert.h"
 #include "dnn_backend_native_layer_mathbinary.h"
 
+float SUB_fun(const float src, float dst)
```
xuguangxin:
```suggestion
static float sub(float src, float dst)
```

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -102,95 +160,23 @@ int dnn_execute_layer_math_binary(DnnOperand *operands, const int32_t *input_ope
...
-            }
-        } else {
-            const DnnOperand *input1 = &operands[input_operand_indexes[1]];
-            const float *src1 = input1->data;
-            for (int i = 0; i < dims_count; ++i) {
-                dst[i] = src[i] - src1[i];
-            }
-        }
+    case DMBO_SUB:  
+        mathbin2(SUB_fun, params, src, dst, dims_count, operands, input_operand_indexes);
```
xuguangxin: if we pass input and output here,
the dims_count is not needed, right?

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -102,95 +160,23 @@ int dnn_execute_layer_math_binary(DnnOperand *operands, const int32_t *input_ope
...
-            }
-        } else {
-            const DnnOperand *input1 = &operands[input_operand_indexes[1]];
-            const float *src1 = input1->data;
-            for (int i = 0; i < dims_count; ++i) {
-                dst[i] = src[i] + src1[i];
-            }
-        }
+    case DMBO_ADD: 
+        mathbin1(ADD_fun, params, src, dst, dims_count, operands, input_operand_indexes);
```
xuguangxin: why we use different function for sub and add?

libavfilter/dnn/dnn_backend_native_layer_mathbinary.h
```diff
@@ -47,6 +47,17 @@ typedef struct DnnLayerMathBinaryParams{
     float v;
 } DnnLayerMathBinaryParams;
 
+typedef float (*FunType)(const float src, float dst);
```
xuguangxin: it's not need be in header, since it's only need by the c file

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: Thanks, it has been modified in the new patch.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: This suggestion has been adopted in the new patch.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: Different from add, sub, realdiv and floormod needs to consider the order of two input data.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.h
MingyuY: It has been modified in the new patch.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
xuguangxin: but, for current code,  add and sub uses different function, right?
>Different from add, sub, realdiv and floormod needs to consider the order of two input data.

Not familiar with this. But tf's realdiv has no order. https://www.w3cschool.cn/tensorflow_python/tensorflow_python-fi9u2hcl.html

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -27,6 +27,79 @@
 #include "libavutil/avassert.h"
 #include "dnn_backend_native_layer_mathbinary.h"
 
+typedef float (*FunType)(const float src, float dst);
+FunType Pfun;
```
guoyejun: pfun

libavfilter/dnn/dnn_backend_native_layer_mathbinary.h
```diff
@@ -47,6 +47,7 @@ typedef struct DnnLayerMathBinaryParams{
     float v;
 } DnnLayerMathBinaryParams;
 
+
 int dnn_load_layer_math_binary(Layer *layer, AVIOContext *model_file_context, int file_size, int operands_num);
```
guoyejun: don't add extra empty line

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -27,6 +27,79 @@
...
+{
+    return FFMIN(src, dst);
+}
+static float floormod(float src, float dst)
+{
+    return (float)((int)(src) % (int)(dst));
+}
+
+static void mathbin1(FunType Pfun, const DnnLayerMathBinaryParams *params, const DnnOperand *input, DnnOperand *output, DnnOperand *operands, const int32_t *input_operand_indexes)
+{
```
guoyejun, line 59: pfun, same for mathbin2.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -27,6 +27,79 @@
...
+static float minimum(float src, float dst)
+{
+    return FFMIN(src, dst);
+}
+static float floormod(float src, float dst)
+{
+    return (float)((int)(src) % (int)(dst));
+}
+
+static void mathbin1(FunType Pfun, const DnnLayerMathBinaryParams *params, const DnnOperand *input, DnnOperand *output, DnnOperand *operands, const int32_t *input_operand_indexes)
```
guoyejun: use a meaningful name for mathbin1 and mathbin2

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -27,6 +27,79 @@
...
+    const float *src;
+    float *dst;
+    dims_count = calculate_operand_dims_count(output);
+    src = input->data;
+    dst = output->data;
+    if (params->input0_broadcast || params->input1_broadcast) {
+            for (int i = 0; i < dims_count; ++i) {
+                dst[i] = Pfun(params->v, src[i]);
+            }
+        } else {
```
guoyejun: indent not right?

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -27,6 +27,79 @@
...
+    const float *src;
+    float *dst;
+    dims_count = calculate_operand_dims_count(output);
+    src = input->data;
+    dst = output->data;
+    if (params->input0_broadcast) {
+            for (int i = 0; i < dims_count; ++i) {
+                dst[i] = Pfun(params->v, src[i]);
+            }
+        } else if (params->input1_broadcast) {
```
guoyejun: indent?

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
```diff
@@ -97,100 +167,24 @@ int dnn_execute_layer_math_binary(DnnOperand *operands, const int32_t *input_ope
...
-                dst[i] = (int)(src[i]) % (int)(params->v);
-            }
-        } else {
-            const DnnOperand *input1 = &operands[input_operand_indexes[1]];
-            const float *src1 = input1->data;
-            for (int i = 0; i < dims_count; ++i) {
-                dst[i] = (int)(src[i]) % (int)(src1[i]);
-            }
-        }
+        mathbin2(floormod, params, input, output, operands, input_operand_indexes);
```
guoyejun: use this style:
case A:
case B:
case C:
   fun1();
   break;
case D:
case E:
   fun2();
   break;
default:
...

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: > but, for current code, add and sub uses different function, right?
> 
> > Different from add, sub, realdiv and floormod needs to consider the order of two input data.
> 
> Not familiar with this. But tf's realdiv has no order. https://www.w3cschool.cn/tensorflow_python/tensorflow_python-fi9u2hcl.html

yes, sub needs to consider the order, and add does not.

MingyuY: > > but, for current code, add and sub uses different function, right?
> > > Different from add, sub, realdiv and floormod needs to consider the order of two input data.
> > 
> > 
> > Not familiar with this. But tf's realdiv has no order. https://www.w3cschool.cn/tensorflow_python/tensorflow_python-fi9u2hcl.html
> 
> This function takes into account the existence of constants in the input. Because when there is a constant, we do not need to traverse every position of the input data for the input constant, so in this case, he will use input0_ broadcast， input1_broadcast represents the input position of the constant, and params->v is the value of the constant. In this case, there are two different orders for sub functions.
If there is no constant, input0_broadcast, input1_broadcast are both 0, then for sub and add, there is no need to consider the order problem, only need to execute the function according to the input order.

MingyuY: > This function takes into account the existence of constants in the input. Because when there is a constant, we do not need to traverse every position of the input data for the input constant, so in this case, he will use input0_ broadcast， input1_broadcast represents the input position of the constant, and params->v is the value of the constant. In this case, there are two different orders for sub functions.
> If there is no constant, input0_broadcast, input1_broadcast are both 0, then for sub and add, there is no need to consider the order problem, only need to execute the function according to the input order.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: > pfun

OK, has been adopted.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.h
MingyuY: > don't add extra empty line

Has been corrected.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: > pfun, same for mathbin2.

Has been adopted.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: > use a meaningful name for mathbin1 and mathbin2

Has been adopted.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: > indent not right?

Thanks，has been corrected.

libavfilter/dnn/dnn_backend_native_layer_mathbinary.c
MingyuY: > use this style:
> case A:
> case B:
> case C:
> fun1();
> break;
> case D:
> case E:
> fun2();
> break;
> default:
> ...

I try it, and the output seems wrong.

guoyejun: i made a mistake, it does not work.

PR #272, opened 2020-08-26
dnn/openvino: add log error message

Semmer2:


libavfilter/dnn/dnn_backend_openvino.h
```diff
@@ -28,6 +28,21 @@
...
+} OVContext;
+
+typedef struct OVModel{
+    OVContext ctx;
+    ie_core_t *core;
+    ie_network_t *network;
+    ie_executable_network_t *exe_network;
+    ie_infer_request_t *infer_request;
+    ie_blob_t *input_blob;
+} OVModel;
```
guoyejun, line 45: why  we need to define ovcontext and ovmodel in this .h file?

Semmer2: Well, I think all the define of struct should be in header file, just like the define of NativeModel & NativeContext in dnn_backend_native.h. Would this be the reason why all CI failed?

guoyejun: no, we should put only necessary things in header file.

looks it has no relation with ci failure.

Semmer2: OK, got you, I will modify this part in next version.

PR #275, opened 2020-08-27
dnn/tensorflow: add log error message

Semmer2:


uartie:
checks passed.  submitted upstream. close.

PR #276, opened 2020-08-28
Fix qsv device

xhaihao:


uartie:
checks passed.  submitted upstream. close.

PR #284, opened 2020-09-04
dnn/openvino: support run inference via GPU

Semmer2:
use func av_opt_set_from_string to find the .options of OVContext.AVClass.options automatically, and report error if not match option found.
if found proper option, it will assign value to OVContext.device_type.
and the verify if the assigned device_type available or not.

### DIFF
--- a/libavfilter/dnn/dnn_backend_openvino.c
+++ b/libavfilter/dnn/dnn_backend_openvino.c
@@ -26,10 +26,18 @@
 #include "dnn_backend_openvino.h"
 #include "libavformat/avio.h"
 #include "libavutil/avassert.h"
+#include "libavutil/opt.h"
+#include "libavutil/avstring.h"
+#include "../internal.h"
 #include <c_api/ie_c_api.h>
+
+typedef struct OVOptions{
+    char *device_type;
+} OVOptions;
 
 typedef struct OVContext {
     const AVClass *class;
+    OVOptions options;
 } OVContext;
 
 typedef struct OVModel{
@@ -41,13 +49,14 @@ typedef struct OVModel{
     ie_blob_t *input_blob;
 } OVModel;
 
-static const AVClass dnn_openvino_class = {
-    .class_name = "dnn_openvino",
-    .item_name  = av_default_item_name,
-    .option     = NULL,
-    .version    = LIBAVUTIL_VERSION_INT,
-    .category   = AV_CLASS_CATEGORY_FILTER,
+#define OFFSET(x) offsetof(OVContext, x)
+#define FLAGS AV_OPT_FLAG_FILTERING_PARAM
+static const AVOption dnn_openvino_options[] = {
+    { "device", "device to run model", OFFSET(options.device_type), AV_OPT_TYPE_STRING, { .str = "CPU" }, 0, 0, FLAGS },
+    { NULL }
 };
+
+AVFILTER_DEFINE_CLASS(dnn_openvino);
 
 static DNNDataType precision_to_datatype(precision_e precision)
 {
@@ -159,10 +168,18 @@ err:
 
 DNNModel *ff_dnn_load_model_ov(const char *model_filename, const char *options)
 {
+    char *a_dev_names = NULL;
     DNNModel *model = NULL;
     OVModel *ov_model = NULL;
+    OVContext *ctx = NULL;
     IEStatusCode status;
     ie_config_t config = {NULL, NULL, NULL};
+    ie_available_devices_t *a_dev;
+    a_dev = av_malloc(sizeof(ie_available_devices_t));
+    if (!a_dev){
+        av_log(NULL, AV_LOG_ERROR, "Failed to allocate memory for device info\n");
+        return NULL;
+    }
 
     model = av_malloc(sizeof(DNNModel));
     if (!model){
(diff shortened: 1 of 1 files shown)


libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -41,14 +44,15 @@ typedef struct OVModel{
...
 } OVModel;
 
-static const AVClass dnn_openvino_class = {
-    .class_name = "dnn_openvino",
-    .item_name  = av_default_item_name,
-    .option     = NULL,
-    .version    = LIBAVUTIL_VERSION_INT,
-    .category   = AV_CLASS_CATEGORY_FILTER,
+#define OFFSET(x) offsetof(OVContext, x)
+#define FLAGS AV_OPT_FLAG_FILTERING_PARAM | AV_OPT_FLAG_VIDEO_PARAM
```
guoyejun: do we need AV_OPT_FLAG_VIDEO_PARAM?

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -26,10 +26,13 @@
 #include "dnn_backend_openvino.h"
 #include "libavformat/avio.h"
 #include "libavutil/avassert.h"
+#include "libavutil/opt.h"
+#include "../internal.h"
 #include <c_api/ie_c_api.h>
 
 typedef struct OVContext {
     const AVClass *class;
+    char *device_type;
```
guoyejun: there might be more options, so let's add code like:
struct OVOptions
{
char *device_type;
}

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -41,14 +44,15 @@ typedef struct OVModel{
...
-static const AVClass dnn_openvino_class = {
-    .class_name = "dnn_openvino",
-    .item_name  = av_default_item_name,
-    .option     = NULL,
-    .version    = LIBAVUTIL_VERSION_INT,
-    .category   = AV_CLASS_CATEGORY_FILTER,
+#define OFFSET(x) offsetof(OVContext, x)
+#define FLAGS AV_OPT_FLAG_FILTERING_PARAM | AV_OPT_FLAG_VIDEO_PARAM
+static const AVOption dnn_openvino_options[] = {
+    { "device", "device to run model", OFFSET(device_type), AV_OPT_TYPE_STRING, { .str = NULL }, 0, 0, FLAGS },
```
guoyejun: can we make .str="CPU"?  so we don't need line 219-222

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -61,6 +65,25 @@ static DNNDataType precision_to_datatype(precision_e precision)
     }
 }
 
+static DNNReturnType parse_options(OVContext *dst, const char *src_opt, ie_available_devices_t *a_dev)
```
guoyejun: should not pass a_dev here, we'll add more options, we cannot add a function parameter for each possible option.

for invalid device, we can rely on the return value of ie_core_load_network.  If it returns error, we can then call ie_core_get_available_devices and list all the supported devices.

libavfilter/dnn/dnn_backend_openvino.c
Semmer2: Well, I think we can make .str="CPU". Will modify in next version.

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -61,6 +65,25 @@ static DNNDataType precision_to_datatype(precision_e precision)
...
+{
+    if (av_opt_set_from_string(dst, src_opt, NULL, "=", "&") < 0) {
```
guoyejun: if av_opt_set_from_string is enough, just call it within ff_dnn_load_model_ov, and remove function parse_options

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -173,16 +203,29 @@ DNNModel *ff_dnn_load_model_ov(const char *model_filename, const char *options)
     if (!ov_model)
         goto err;
     ov_model->ctx.class = &dnn_openvino_class;
+    ctx = &ov_model->ctx;
```
guoyejun, line 193: why we need this?  since line 215 still use &ov_model->ctx

libavfilter/dnn/dnn_backend_openvino.c
Semmer2: um, as I know the options are setted in struct XXXContext generally, so the funcs (in dict.c and opt.c) can assign values to membors of XXXContext and show error log info with ctx.
If we create new struct like OVOption, and pass it to funcs(in dict.c & opt.c), the error log info will not be shown correctly.

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -173,16 +203,29 @@ DNNModel *ff_dnn_load_model_ov(const char *model_filename, const char *options)
...
+    }
+
     status = ie_core_read_network(ov_model->core, model_filename, NULL, &ov_model->network);
     if (status != OK)
         goto err;
 
-    status = ie_core_load_network(ov_model->core, ov_model->network, "CPU", &config, &ov_model->exe_network);
+    status = ie_core_load_network(ov_model->core, ov_model->network, ov_model->ctx.device_type, &config, &ov_model->exe_network);
     if (status != OK)
         goto err;
```
guoyejun, line 230: list all the supported devices if status is not OK

libavfilter/dnn/dnn_backend_openvino.c
Semmer2: Sorry, I didn't take this seriously just copy it from vf_dnn_processing.c. Will check it later.

libavfilter/dnn/dnn_backend_openvino.c
Semmer2: I think you mean we can verify if the value of device_type match the available device in ff_dnn_load_model_ov() rather than in parse_option() ?
becasue the return of ie_core_load_network can not reflect if the specified device ready or not. (I have checked the IESatatusCode in OpenVINO api just now, no such value like: DEVICE_NOT_FOUND or NO_SUCH_DEVICE)

libavfilter/dnn/dnn_backend_openvino.c
Semmer2: OK, got you

libavfilter/dnn/dnn_backend_openvino.c
Semmer2: since we need the ctx to show log in following av_log()
which have been missed the the "dnn/openvino: add error log" patch

libavfilter/dnn/dnn_backend_openvino.c
Semmer2: OK, we should list all available device rather than use CPU defaultly, right?
Will modify in next patch.

libavfilter/dnn/dnn_backend_openvino.c
guoyejun: i guess it is possible, to pass OVContext.options.device_type.  The offset is OFFSET(options.device_type) instead of OFFSET(device_type)

libavfilter/dnn/dnn_backend_openvino.c
guoyejun: ok

libavfilter/dnn/dnn_backend_openvino.c
guoyejun: no, the default is "CPU", I mean the code like:

if (status != OK) {
    output message:  status, and list all the supported devices
    goto err;
}

libavfilter/dnn/dnn_backend_openvino.c
guoyejun: just ignore this since we will remove this function,

PR #294, opened 2020-09-14
fate test for dnn_module v4

Avery-xu:


guoyejun (commented):
looks good to me.

libavfilter/dnn/dnn_backend_native_layer_conv2d.c
```diff
@@ -227,13 +224,13 @@ int dnn_execute_layer_conv2d(DnnOperand *operands, const int32_t *input_operand_
     thread_common_param.output_data = output_operand->data;
 
 #if HAVE_PTHREAD_CANCEL
-    thread_common_param.thread_num = thread_num;
-
+    int thread_stride = (height - pad_size * 2) / thread_num;
```
guoyejun: build warning found at https://patchwork.ffmpeg.org/project/ffmpeg/patch/20200914113154.61946-2-xujunzz@sjtu.edu.cn/, please fix by move "int thread_stride" up.

uartie: How does upstream check for new warnings?  We'd like to enable this in our CI pre-check too.  But ffmpeg is already littered with compile warnings, so we can't just enable -Werror.
cc: @FocusLuo @Bin-QA

guoyejun: i don't know how ffmpeg checks it, but i guess it is checked via differing the build log for the "New warnings during build"

uartie: Hmm, so they build it twice (one without patch and one with patch) to get a warning diff.  Seems like a hypocrisy and sloppy way to manage warnings.  It would be better if ffmpeg just fixed all their current warnings and then enforce -Werror instead.

guoyejun: a possible method to avoid the two times build is that we first save the original build log (the gold version) as the compare base.

Bin-QA: It is more like build the original source code, apply patch, build again  
Base on `make` policy: ignore without code change  in build process to speed up build result
So this round report build information just for this patch  
How about this idea ?

guoyejun: the only concern is that if we could 100% trust the increment build.

uartie: We need to avoid incremental builds in CI environment.  Sometimes an incremental build can hide issues.  So, in CI we should always do a full, clean build.

PR #302, opened 2020-10-20
ffmpeg_qsv: repect hwaccel_output_format setting

xhaihao:
Without thix change, the output pixel format from qsv decoder is always
qsv regardless of the hwaccel_output_format setting, user will get
rrors when FFmpeg inserts an auto filter to convert the decoder output
to the specified format.

Fox example:

ffmpeg -loglevel verbose -hwaccel qsv -c:v h264_qsv
-hwaccel_output_format yuv420p -i input.h264
-pix_fmt rgb24 -f null -

......
[graph 0 input from stream 0:0 @ 0x55c057c8f050] w:1280 h:720 pixfmt:qsv
tb:1/1001 fr:30000/1001 sar:1/1
[auto_scaler_0 @ 0x55c057c90d20] w:iw h:ih flags:'bicubic' interl:0
[format @ 0x55c057c8f990] auto-inserting filter 'auto_scaler_0' between
the filter 'Parsed_null_0' and the filter 'format'
Impossible to convert between the formats supported by the filter
'Parsed_null_0' and the filter 'auto_scaler_0'
Error reinitializing filters!
Failed to inject frame into filter network: Function not implemented
Error while processing the decoded data for stream #0:

### DIFF
--- a/fftools/ffmpeg_opt.c
+++ b/fftools/ffmpeg_opt.c
@@ -884,6 +884,12 @@ static void add_input_streams(OptionsContext *o, AVFormatContext *ic)
                     "with old commandlines. This behaviour is DEPRECATED and will be removed "
                     "in the future. Please explicitly set \"-hwaccel_output_format cuda\".\n");
                 ist->hwaccel_output_format = AV_PIX_FMT_CUDA;
+            } else if (!hwaccel_output_format && hwaccel && !strcmp(hwaccel, "qsv")) {
+                av_log(NULL, AV_LOG_WARNING,
+                    "WARNING: defaulting hwaccel_output_format to qsv for compatibility "
+                    "with old commandlines. This behaviour is DEPRECATED and will be removed "
+                    "in the future. Please explicitly set \"-hwaccel_output_format qsv\".\n");
+                ist->hwaccel_output_format = AV_PIX_FMT_QSV;
             } else if (hwaccel_output_format) {
                 ist->hwaccel_output_format = av_get_pix_fmt(hwaccel_output_format);
                 if (ist->hwaccel_output_format == AV_PIX_FMT_NONE) {
--- a/fftools/ffmpeg_qsv.c
+++ b/fftools/ffmpeg_qsv.c
@@ -82,6 +82,8 @@ int qsv_init(AVCodecContext *s)
             return ret;
     }
 
+    hwaccel_decode_init(s);
+
     av_buffer_unref(&ist->hw_frames_ctx);
     ist->hw_frames_ctx = av_hwframe_ctx_alloc(hw_device_ctx);
     if (!ist->hw_frames_ctx)


xhaihao:
After applying this patch, `hwdownload` is not required any more if user doesn't specify output format from qsv decoder, 
Before: 
```
ffmpeg -hwaccel qsv -c:v h264_qsv -i input.mp4 -vf hwdownload,format=nv12 -pix_fmt yuv420p output.yuv

```
After:
```
ffmpeg -hwaccel qsv -c:v h264_qsv -i input.mp4 -pix_fmt yuv420p output.yuv
```
The new command line is similar to the command line by using VAAPI.
```
ffmpeg -hwaccel vaapi -i input.mp4 -pix_fmt yuv420p output.yuv
```
@uartie @FocusLuo Any comment for this patch? We will have to modify our cases if applying this patch, the advantage is we will have similar command lines for both qsv and vaapi when using hwdownload

uartie:
Is it possible to make it backwards compatible to support both old and new syntax?  This would make the transition easier.

xhaihao:
Maybe the answer is no. 

The default `hwaccel_output_format` is AV_PIX_FMT_NONE if user doesn't set `hwaccel_output_format` explicitly. However the current implementation of QSV doesn't respect `hwaccel_output_format` setting. Without this patch, the output pixel format from qsv decoder is alway `qsv`, which can not be converted to a software pixel format directly, hence `hwdownload` is needed in the command.

So we will change the default output pixel format from qsv decoder if we want to respect `hwaccel_output_format` setting, however `hwdownload` doesn't support software pixel format input, we have to remove `hwdownload` from the command.

After applying the patch, you may specify `hwaccel_output_format` to qsv if you want to use `hwdownload` 

```
ffmpeg -hwaccel qsv -c:v h264_qsv -hwaccel_output_format qsv -i input.mp4 -vf hwdownload,format=nv12 -pix_fmt yuv420p output.yuv
```

uartie:
Ok, we can add `-hwaccel_output_format qsv` to decode test command lines so that they will work with or without this patch.  Then, once this patch is merged upstream we can simplify the test command lines afterwards.

uartie:
See test suite update: https://github.com/intel/vaapi-fits/pull/239

xhaihao:
I found a way not to change the default behavior if user doesn't set `hwaccel_output_format` explicitly, so we needn't change our test cases. The disadvantage is that qsv and vaapi have different behavior, user would be easily confused on it.  We may change the behavior of qsv in the future.

uartie:
I already merged the test case changes earlier.

xhaihao:
Yes, I saw your change. The new version works with your change, but keep the behavior unchanged and `hwdownload` is needed if user doesn't set `hwaccel_output_format` explicitly.

xhaihao:
My concern is we may change our test cases easily, but for external user, maybe they are not aware of the change of the behavior in the old version,  hence I prefer not to change the default behavior in this patch and added a warning message about the default behavior.

uartie:
> My concern is we may change our test cases easily, but for external user, maybe they are not aware of the change of the behavior in the old version, hence I prefer not to change the default behavior in this patch and added a warning message about the default behavior.

Yes, I agree.  Let's not surprise and confuse the user 😃

xhaihao:
checks passed.

PR #303, opened 2020-11-05
qsvenc_hevc: allow user set more coding options

xhaihao:
The SDK supports NalHrdConformance, RecoveryPointSEI and AUDelimiter for
hevc encoder, so we may allow user to set these coding options like as
what we did for h264_qsv encoder.


xhaihao:
passed.

PR #304, opened 2020-11-05
qsvenc: Value '0' is not a valid value for parameter GopOptFlag

xhaihao:
The accepted values for GopOptFlag are MFX_GOP_CLOSED (1) and
MFX_GOP_STRICT (2).


xhaihao:
passed,

PR #306, opened 2020-11-17
lavf/utils: return NULL if no proper decoder probed

feiwan1:
If all candiate codecs has capability of AV_CODEC_CAP_AVOID_PROBING,
return NULL.


uartie:
@feiwan1 closing.  please resubmit when you fix the tests

PR #350, opened 2021-02-02
libavutil/hwcontext_qsv.c:Fill pixels for realigned frame in qsv

zhangyuankun-star:
According to MediaSDK doc, app needs to fill the extend area caused by alignment,
So I fill the copy of pixel on the border.
MediaSDK doc: https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#encoding-procedures,
"It is the application's responsibility to fill pixels outside of crop window when it is smaller than frame to be encoded.
Especially in cases when crops are not aligned to minimum coding block size (16 for AVC, 8 for HEVC and VP9)."

Signed-off-by: Zhang Yuankun <yuankunx.zhang@intel.com>

### DIFF
--- a/libavutil/hwcontext_qsv.c
+++ b/libavutil/hwcontext_qsv.c
@@ -876,6 +876,50 @@ static int qsv_transfer_data_from(AVHWFramesContext *ctx, AVFrame *dst,
     return 0;
 }
 
+static int qsv_extend_data_for_realigned(AVFrame *dst, const AVFrame *src)
+{
+    int padx, pady, width, height;
+    int plane, offset, step, stride;
+    int i, r, x;
+    uint8_t *pData;
+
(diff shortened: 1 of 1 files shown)


zhangyuankun-star:
@xhaihao @xuguangxin Hi，Haihao & Guangxin,can you review this for me?

xhaihao:
Extending data is required for encoding only, It seems your patch will impact other QSV operations.

uartie:
@zhangyuankun-star any update?

uartie:
@xhaihao @xuguangxin can you follow-up with @zhangyuankun-star about this patch... there has been no response from @zhangyuankun-star

zhangyuankun-star:
I'm sorry I didn't reply in time. I haven't come up with a good solution to the problem raised by haihao. I'll close this PR first, and I'll submit it again after I solve the problem. thanks！

PR #352, opened 2021-02-08
libavcodec/qsvdec: skip non-key frame after "seek" function

wenbinc-Bin:
Fix #9095. Qsv decoder assume that after calling seek funcion, the first
frame should be key frame. However this is not true for some videos. If
the frame is not key frame after seek(), there will error. Conditional
statements are added to skip these frame until reading a key frame.

@xhaihao @xuguangxin @feiwan1 please help me to review.

thanks

### DIFF
--- a/libavcodec/qsvdec.c
+++ b/libavcodec/qsvdec.c
@@ -616,6 +616,13 @@ int ff_qsv_process_data(AVCodecContext *avctx, QSVContext *q,
     }
 
     if (!q->initialized) {
+        
+        /*  skip non-key frame when decoder is reinitialized. 
+        Adding before will skip all the non-key frames so add 
+        it here */
+        if (ret < 0)
+            return ret;
+
         ret = qsv_decode_init_context(avctx, q, &param);
         if (ret < 0)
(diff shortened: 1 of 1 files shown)


libavcodec/qsvdec.c
```diff
@@ -616,6 +616,8 @@ int ff_qsv_process_data(AVCodecContext *avctx, QSVContext *q,
     }
 
     if (!q->initialized) {
+        if (ret < 0)
```
xuguangxin, line 623: is it possible to move  this to line 616?

wenbinc-Bin: I think no. If this is moved out of this conditional statement.  It will skip all the non-key frames all the time. It only need to skip frames when it reinitialize decoder.

xuguangxin: better add some comment here, explain why we not return before this

wenbinc-Bin: Ok

wenbinc-Bin:
submit to community.

PR #356, opened 2021-03-02
Add draw dnn bbox

Semmer2:
The latest 3 commits are my patches. Others are Yejun's patch, which have not been merged.


Semmer2:
Wrong PR, please ignore, will create the correct PR later.

libavfilter/vf_drawbox.c
```diff
@@ -81,6 +82,7 @@ typedef struct DrawBoxContext {
     char *t_expr;          ///< expression for thickness
     int have_alpha;
     int replace;
+    int side_data;
```
guoyejun, line 85: use another name such as 'box_from' or 'box_source' or 'box_sidedata', and the value may be of type AVFrameSideDataType, so we can draw box for boudingbox or AV_FRAME_DATA_REGIONS_OF_INTEREST.

btw, i have a v5 patch just now, it does not use private_ref, just wait for community's comment.

libavfilter/vf_drawbox.c
```diff
@@ -217,58 +219,85 @@ static av_pure av_always_inline int pixel_belongs_to_box(DrawBoxContext *s, int
...
+        if (bbox) {
+        s->y = bbox->top;
+        s->x = bbox->left;
+        s->h = bbox->bottom - bbox->top;
+        s->w = bbox->right - bbox->left;
+        xb = s->x;
+        yb = s->y;
+        }
+
+        if (s->have_alpha && s->replace) {
```
guoyejun, line 249: we'd better first extract an function such as static void drawbox(frame, s, left, top, width, height) in a separate patch, and then, call this function in a loop to draw multiple bounding boxes.

PR #358, opened 2021-03-03
libavcodec/vaapi_encode: fix #7706. Call vaapi encoder func to async way

wenbinc-Bin:
After commit 5fdcf85bbffe7451c2, vaapi encoder's performance drop
20~30%. One reason is that vaRenderPicture() and vaSyncSurface() are
called at the same time (vaRenderPicture() always followed by a
vaSyncSurface()). Now I changed them to be called in a
asynchronous way, which will make better use of hardware.

Another reason of performance drop is that in old version, ffmpeg-vaapi
use CQP as default while the ffmpeg-vaapi of lastest version does not,
so the same command line will have a even bigger performance gap. To
test this patch you'd better specify the bitrate (-b:v XXX).

Signed-off-by: Wenbin CHEN <wenbin.chen@intel.com>

### DIFF
--- a/libavcodec/vaapi_encode.c
+++ b/libavcodec/vaapi_encode.c
@@ -950,8 +950,10 @@ static int vaapi_encode_pick_next(AVCodecContext *avctx,
     if (!pic && ctx->end_of_stream) {
         --b_counter;
         pic = ctx->pic_end;
-        if (pic->encode_issued)
+        if (pic->encode_complete)
             return AVERROR_EOF;
+        else if (pic->encode_issued)
+            return AVERROR(EAGAIN);
     }
 
     if (!pic) {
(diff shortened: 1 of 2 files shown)


wenbinc-Bin:
@xhaihao @xuguangxin @feiwan1 .
Hi, please help me to review this patch. 
Thans very much.

xuguangxin:
This will introduce encode delay...

wenbinc-Bin:
@xuguangxin 
Hi guangxin. I don't think this patch introduces encoding delay, because what encoder receive and send at each time are as the same as before. The difference is that instead of sending only one frame to libva it also send following frames to libva, because sometimes we know what frames will be encoded, so that next time encoder can directly call vaSyncSurface() to get encoded frame. The fifo doesn't act as buffer and it is used to record the order of frames that are sent to libva.
Thanks for review
wenbin

wenbinc-Bin:
Discussed with guangxin offline, and will submit to community.

PR #359, opened 2021-03-09
libavcodec/qsvdec: change the get surface logic

wenbinc-Bin:
MSDK vc1 and av1 sometimes output frame into the same suface, but
ffmpeg-qsv assume the surface will be used only once, so it will
unref the frame when it receives the outpur surface. Now change
it to allocate frame at runtime but unref when whole decoding process
end.

Signed-off-by Wenbin Chen <wenbin.chen@intel.com>

### DIFF
--- a/libavcodec/qsvdec.c
+++ b/libavcodec/qsvdec.c
@@ -513,7 +513,7 @@ static int qsv_decode(AVCodecContext *avctx, QSVContext *q,
             return AVERROR_BUG;
         }
 
-        out_frame->queued = 1;
+        out_frame->queued += 1;
         av_fifo_generic_write(q->async_fifo, &out_frame, sizeof(out_frame), NULL);
         av_fifo_generic_write(q->async_fifo, &sync,      sizeof(sync),      NULL);
     } else {
@@ -526,7 +526,7 @@ static int qsv_decode(AVCodecContext *avctx, QSVContext *q,
 
         av_fifo_generic_read(q->async_fifo, &out_frame, sizeof(out_frame), NULL);
         av_fifo_generic_read(q->async_fifo, &sync,      sizeof(sync),      NULL);
-        out_frame->queued = 0;
+        out_frame->queued -= 1;
 
         if (avctx->pix_fmt != AV_PIX_FMT_QSV) {
             do {


uartie:
Interesting.  Will this fix the VC1 R2R issue?

uartie:
> Interesting. Will this fix the VC1 R2R issue?

Ha, I see your branch name is vc1-r2r... so yes! 😄 

When you submit this to ML, can you also submit to cartwheel-ffmpeg, please?

wenbinc-Bin:
Ok, I will submit to cartwheel-ffmpeg too. I found the root cause but there is still some bug to fix for my patch.

wenbinc-Bin:
I will submit it after it pass the test. By the way, what do you mean by ML?

wenbinc-Bin:
@xuguangxin @xhaihao @feiwan1 could you please help me review this patch.
Thanks very much.

uartie:
> I will submit it after it pass the test. By the way, what do you mean by ML?

ML = Mailing List (i.e. the upstream ffmpeg mailing list)

wenbinc-Bin:
Ok, got it. Thanks for you explanation.

xuguangxin:
Hi @wenbinc-Bin , 
Could we attach some public clips in commit log? It will help community reproduce the issue.
Is this PR based on your previous patches? If yes, maybe we need ask community review and push your previous patches firstly.

libavcodec/qsvdec.c
```diff
@@ -340,7 +340,6 @@ static void qsv_clear_unused_frames(QSVContext *q)
     while (cur) {
         if (cur->used && !cur->surface.Data.Locked && !cur->queued) {
             cur->used = 0;
-            av_frame_unref(cur->frame);
```
xuguangxin: Downstream may av_frame_ref the cur->frame and put it to encoder reorder queue. You can not reuse for decode output.

wenbinc-Bin:
> Hi @wenbinc-Bin ,
> Could we attach some public clips in commit log? It will help community reproduce the issue.
> Is this PR based on your previous patches? If yes, maybe we need ask community review and push your previous patches firstly.

I have a video clips can produce error but I don't know whether it can be public. This patch doesn't based on my previous patches.

libavcodec/qsvdec.c
wenbinc-Bin: You are right. There will be problem even if it passes the test.
I found a very simple way to solve the r2r problem and I have pushed to this branch again. You can review the new one.

xuguangxin:
> > Hi @wenbinc-Bin ,
> > Could we attach some public clips in commit log? It will help community reproduce the issue.
> > Is this PR based on your previous patches? If yes, maybe we need ask community review and push your previous patches firstly.
> 
> I have a video clips can produce error but I don't know whether it can be public. This patch doesn't based on my previous patches.

you can send the name to me using mail. I will check.

xuguangxin (approved):
Great! it's a good solution. 
If no objections, we can merge it and check what's happened in full round.
Not sure gstreamer has r2r issue or not. If yes. it maybe the same reason.

wenbinc-Bin:
ok, I will submit patch to community and then submit it to cartwheel.

PR #361, opened 2021-03-15
lavfi/qsvvpp: support async depth

feiwan1:
Async depth will allow qsv filter cache few frames, and avoid force
switch and end filter task frame by frame. This change will improve
performance for some multi-task case, for example 1:N transcode(
decode + vpp + encode) with all QSV plugins.


feiwan1:
Is this a test platform issue? I tested the 1st failed case on my CFL, it looks good. And all other fail case test cmd is empty.

PR #375, opened 2021-03-25
Dnn detect tensorflow support

Semmer2:
Support dnn detect only, do not support classify for now


libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -804,6 +805,21 @@ static DNNReturnType execute_model_tf(const DNNModel *model, const char *input_n
             out_frame->width = output[0].width;
             out_frame->height = output[0].height;
         }
+        break;
+    case DFT_ANALYTICS_DETECT:
+        if (!model->detect_post_proc) {
+            av_log(ctx, AV_LOG_ERROR, "detect filter needs to provide post proc\n");
+            return DNN_ERROR;
+        }
+        model->detect_post_proc(out_frame, output, nb_output, model->filter_ctx);
```
guoyejun, line 814: does your detection work without this patch?

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -153,7 +153,7 @@ static DNNReturnType get_input_tf(void *model, DNNData *input, const char *input
     TF_DeleteStatus(status);
 
     // currently only NHWC is supported
-    av_assert0(dims[0] == 1);
+    av_assert0(dims[0] == 1 || dims[0] == -1);
```
guoyejun, line 157: is this change enough to "support model with random input"?

btw, the title 'random input' is not clear.

libavfilter/dnn/dnn_io_proc.c
```diff
@@ -168,8 +168,12 @@ static DNNReturnType proc_from_frame_to_dnn_frameprocessing(AVFrame *frame, DNND
...
 static enum AVPixelFormat get_pixel_format(DNNData *data)
 {
-    if (data->dt == DNN_UINT8 && data->order == DCO_BGR) {
-        return AV_PIX_FMT_BGR24;
+    if (data->dt == DNN_UINT8) {
+        if (data->order == DCO_BGR) {
+            return AV_PIX_FMT_BGR24;
+        } else if (data->order == DCO_RGB) {
+            return AV_PIX_FMT_RGB24;
+        }
```
guoyejun, line 179: use switch/case, not if/else

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -790,22 +784,25 @@ static DNNReturnType execute_model_tf(const DNNModel *model, const char *input_n
...
+    }
+    {
+        //it only support 1 output if it's frame in & frame out
         if (do_ioproc) {
             if (tf_model->model->frame_post_proc != NULL) {
-                tf_model->model->frame_post_proc(out_frame, &output, tf_model->model->filter_ctx);
+                tf_model->model->frame_post_proc(out_frame, output, tf_model->model->filter_ctx);
             } else {
-                ff_proc_from_dnn_to_frame(out_frame, &output, ctx);
+                ff_proc_from_dnn_to_frame(out_frame, &output[0], ctx);
```
guoyejun, line 802: callback for detection missed

libavfilter/dnn_filter_common.h
```diff
@@ -51,6 +53,7 @@ int ff_dnn_init(DnnContext *ctx, DNNFunctionType func_type, AVFilterContext *fil
 int ff_dnn_set_frame_proc(DnnContext *ctx, FramePrePostProc pre_proc, FramePrePostProc post_proc);
 int ff_dnn_set_detect_post_proc(DnnContext *ctx, DetectPostProc post_proc);
 int ff_dnn_set_classify_post_proc(DnnContext *ctx, ClassifyPostProc post_proc);
+char **ff_dnn_parse_values(DnnContext *ctx, const char *output_exp, const char *val_sep, int *nb_output);
```
guoyejun: what does this function do?  We need a meaningful function name.

libavfilter/dnn_filter_common.c
```diff
@@ -19,6 +19,44 @@
...
+        if (output_nb != 1) {
+            av_log(ctx, AV_LOG_ERROR, "Dnn processing filter needs 1 output only, \
+                                       but get %d instead\n", output_nb);
+            return AVERROR(EINVAL);
+        }
+        return 0;
+    case DFT_ANALYTICS_DETECT:
+        switch(backend_type) {
+        case DNN_TF:
+            if (output_nb != 4) {
```
guoyejun: this logic is special in vf_detect.c for tf backend, why we check here?

libavfilter/vf_dnn_detect.c
```diff
@@ -62,29 +62,30 @@ static const AVOption dnn_detect_options[] = {
 
 AVFILTER_DEFINE_CLASS(dnn_detect);
 
-static int dnn_detect_post_proc(AVFrame *frame, DNNData *output, uint32_t nb, AVFilterContext *filter_ctx)
+static int parse_output_to_bounding_box(AVFilterContext *filter_ctx, AVFrame *frame,
+                                        const int position_size,     const int other_size,
+                                        const float *conf_p,         const float *label_id_p,
+                                        const float *position_p,     const int proposal_count,
+                                        const int x0_offset,         const int y0_offset,
+                                        const int x1_offset,         const int y1_offset)
```
guoyejun: parse_output_to_bounding_box is designed as a general function, but it would be very complex when more models are supported, for example, some model might not output confidence, etc.

so, we can rename the original dnn_detect_post_proc to dnn_detect_post_proc_ov, and add a new function dnn_detect_post_proc_tf. Keep them separated.

libavfilter/dnn/dnn_backend_tf.c
Semmer2: Yes, I tested tf detection with a picture of dog, and it works well with all my patches applied.

libavfilter/dnn/dnn_io_proc.c
Semmer2: OK

libavfilter/dnn_filter_common.h
Semmer2: To parse multiple outputs, with '&' to seperate. Maybe parse_output_name_expr better?
While, i thought may be there would be other condition, maybe this func would be called to parse multiple names/expressions with some certain seperation symbol. So i choosed a more common func name.

libavfilter/dnn_filter_common.c
Semmer2: I think maybe tf classify model may reuse this logic code in following code. Since tf calssify model needs multiple outputs as well.

libavfilter/dnn/dnn_backend_tf.c
Semmer2: Just for fixing the error of "input dims[0] != 1". Maybe i should change this commit message to something like: dnn/dnn_backend_tf: fix error of model input batch_size = -1 ?

libavfilter/dnn/dnn_backend_tf.c
Semmer2: Yes, this patch does not work well, detect callback func is in the last patch. Maybe current patch set is not resonable, i will think again.

libavfilter/dnn/dnn_backend_tf.c
guoyejun: dims[0] is batch size, we'd put all the changes relative to batch size together as a single patch

libavfilter/dnn/dnn_backend_tf.c
guoyejun: we need to merge part of this change into others, and in the last patch to enable tf+detect, we'd also add detail command lines.

libavfilter/dnn/dnn_backend_tf.c
guoyejun: yes, thanks.

libavfilter/dnn_filter_common.h
guoyejun: we still need a meaningful name even it is common, for example, separate_expressions or something else.

libavfilter/dnn_filter_common.c
guoyejun: the magic number 4 should be special

libavfilter/dnn/dnn_backend_tf.c
Semmer2: OK, sure

libavfilter/dnn/dnn_backend_tf.c
Semmer2: OK

libavfilter/dnn_filter_common.h
Semmer2: Yes, i get it

libavfilter/dnn_filter_common.c
Semmer2: OK~ So is it ok that i put the logic here?

libavfilter/vf_dnn_detect.c
Semmer2: Oh, you're right, but these two func have so many common codes, maybe I should use a macro? Or just leave it?

guoyejun: we can just leave it unless there's other comment

libavfilter/dnn_filter_common.c
guoyejun: just put in vf_dnn_detect.c now, we can refine the code if it is common later

Semmer2: OK, got you

libavfilter/vf_dnn_detect.c
Semmer2: OK

libavfilter/vf_dnn_detect.c
```diff
@@ -211,7 +214,12 @@ static int read_detect_label_file(AVFilterContext *context)
 static av_cold int dnn_detect_init(AVFilterContext *context)
 {
     DnnDetectContext *ctx = context->priv;
-    int ret = ff_dnn_init(&ctx->dnnctx, DFT_ANALYTICS_DETECT, context);
+    DnnContext *dnn_ctx = &ctx->dnnctx;
+    int ret;
+
+    dnn_ctx->model_outputname = ff_dnn_separate_expression(dnn_ctx, dnn_ctx->outputname_expr,
```
guoyejun: better to be done within ff_dnn_init, in other words, don't export function ff_dnn_separate_expression

libavfilter/dnn_filter_common.h
```diff
@@ -30,7 +30,9 @@ typedef struct DnnContext {
     char *model_filename;
     DNNBackendType backend_type;
     char *model_inputname;
-    char *model_outputname;
+    char **model_outputname;
```
guoyejun: model_output_names

libavfilter/vf_dnn_detect.c
```diff
@@ -48,6 +48,9 @@ typedef struct DnnDetectContext {
 #define FLAGS AV_OPT_FLAG_FILTERING_PARAM | AV_OPT_FLAG_VIDEO_PARAM
 static const AVOption dnn_detect_options[] = {
     { "dnn_backend", "DNN backend",                OFFSET(backend_type),     AV_OPT_TYPE_INT,       { .i64 = 2 },    INT_MIN, INT_MAX, FLAGS, "backend" },
+#if (CONFIG_LIBTENSORFLOW == 1)
+    { "tensorflow",  "tensorflow backend flag",    0,                        AV_OPT_TYPE_CONST,     { .i64 = 1 },    0, 0, FLAGS, "backend" },
+#endif
```
guoyejun, line 53: if you enable this option in this patch, the tf backend should work here

libavfilter/vf_dnn_processing.c
```diff
@@ -62,6 +62,9 @@ AVFILTER_DEFINE_CLASS(dnn_processing);
 static av_cold int init(AVFilterContext *context)
 {
     DnnProcessingContext *ctx = context->priv;
+    DnnContext *dnn_ctx = &ctx->dnnctx;
+    dnn_ctx->model_outputname = ff_dnn_separate_expression(dnn_ctx, dnn_ctx->outputname_expr,
+                                                           "&", &dnn_ctx->nb_output);
```
guoyejun: this patch shows that: dnn_processing is broken by your previous patch, it is not allowed. Please think carefully about how to split/merge the patches.

libavfilter/vf_dnn_detect.c
```diff
@@ -211,6 +211,30 @@ static int read_detect_label_file(AVFilterContext *context)
     return 0;
 }
 
+static int check_output_nb(DnnDetectContext *ctx, DNNBackendType backend_type, int output_nb)
+{
+    switch(backend_type) {
+    case DNN_TF:
+        if (output_nb != 4) {
+            av_log(ctx, AV_LOG_ERROR, "Dnn detect filter with tensorflow backend needs 4 outputs, \
```
guoyejun: it is not tf backend needs 4 outputs, but that we currently only support models with 4 outputs.

libavfilter/vf_dnn_detect.c
```diff
@@ -62,9 +62,10 @@ static const AVOption dnn_detect_options[] = {
 
 AVFILTER_DEFINE_CLASS(dnn_detect);
 
-static int dnn_detect_post_proc(AVFrame *frame, DNNData *output, uint32_t nb, AVFilterContext *filter_ctx)
+static int dnn_detect_post_proc_ov(AVFrame *frame, DNNData *output, uint32_t nb, AVFilterContext *filter_ctx)
 {
     DnnDetectContext *ctx = filter_ctx->priv;
+    float x0, y0, x1, y1;
```
guoyejun: don't change the code that is not relative to your patch

libavfilter/vf_dnn_detect.c
```diff
@@ -95,6 +96,7 @@ static int dnn_detect_post_proc(AVFrame *frame, DNNData *output, uint32_t nb, AV
 
     sd = av_frame_new_side_data(frame, AV_FRAME_DATA_BOUNDING_BOXES,
                                 sizeof(*header) + sizeof(*bbox) * nb_bbox);
+
```
guoyejun: don't change the code that is not relative to your patch

libavfilter/vf_dnn_detect.c
```diff
@@ -249,7 +332,17 @@ static av_cold int dnn_detect_init(AVFilterContext *context)
     ret = ff_dnn_init(&ctx->dnnctx, DFT_ANALYTICS_DETECT, context);
     if (ret < 0)
         return ret;
-    ff_dnn_set_detect_post_proc(&ctx->dnnctx, dnn_detect_post_proc);
```
guoyejun: still use dnn_detect_post_proc as a single entry, but to go to different path within this function.

libavfilter/vf_drawbox.c
```diff
@@ -389,58 +397,8 @@ static av_pure av_always_inline int pixel_belongs_to_grid(DrawBoxContext *drawgr
...
-                    if (pixel_belongs_to_grid(drawgrid, x, y)) {
-                        row[0][x                  ] = (1 - alpha) * row[0][x                  ] + alpha * drawgrid->yuv_color[Y];
-                        row[1][x >> drawgrid->hsub] = (1 - alpha) * row[1][x >> drawgrid->hsub] + alpha * drawgrid->yuv_color[U];
-                        row[2][x >> drawgrid->hsub] = (1 - alpha) * row[2][x >> drawgrid->hsub] + alpha * drawgrid->yuv_color[V];
-                    }
-                }
-            }
-        }
-    }
+    draw_box_and_grid(frame, drawgrid, 0, 0, frame->width, frame->height, pixel_belongs_to_grid);
```
guoyejun: please just remove the patches about vf_drawbox/text in this patch set, let's only focus on detect+tf here

guoyejun:
basically, you can merge them into two patches, one for the tf backend, the other for the detect filter.  And 1-2 more small patches is possible if you think some changes in tf backend is not special for detect.

guoyejun:
btw, please rebase your patches on my detect patch (do not include classify), so your patches can be merged after my detect is pushed.

Semmer2:
OK, got yout

PR #378, opened 2021-03-31
qsv: fix the number of threads created by a mfx session

xhaihao:
I tried to reduce the number of threads created by the SDK when using FFmpeg-QSV

### DIFF
--- a/libavcodec/qsv.c
+++ b/libavcodec/qsv.c
@@ -390,11 +390,27 @@ int ff_qsv_init_internal_session(AVCodecContext *avctx, QSVSession *qs,
     const char *desc;
     int ret;
 
+#if QSV_VERSION_ATLEAST(1, 15)
+    mfxExtBuffer *ext_params[1];
+    mfxExtThreadsParam thread_param;
+#endif
+
 #if QSV_VERSION_ATLEAST(1, 16)
     init_par.GPUCopy        = gpu_copy;
 #endif
     init_par.Implementation = impl;
     init_par.Version        = ver;
+
+#if QSV_VERSION_ATLEAST(1, 15)
+    memset(&thread_param, 0, sizeof(thread_param));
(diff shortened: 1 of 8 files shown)


uartie:
LGTM

libavcodec/qsv.c
```diff
@@ -390,11 +390,27 @@ int ff_qsv_init_internal_session(AVCodecContext *avctx, QSVSession *qs,
...
+
+#if QSV_VERSION_ATLEAST(1, 15)
+    memset(&thread_param, 0, sizeof(thread_param));
+    thread_param.Header.BufferId = MFX_EXTBUFF_THREADS_PARAM;
+    thread_param.Header.BufferSz = sizeof(thread_param);
+    thread_param.NumThread       = FFMAX(2, avctx->thread_count);
+    ext_params[0]                = (mfxExtBuffer *)&thread_param;
+    init_par.ExtParam            = (mfxExtBuffer **)&ext_params;
+    init_par.NumExtParam         = 1;
+#endif
```
xuguangxin, line 412: duplicate code could we use one function for this?
qsv_init_mfx(init_par, session, maxthreads)

xhaihao: libavcodec and libavutils are two different libraries. It seems the FFmpeg community doesn't prefer avpriv_ functions (which are used  internally across multiple libraries) and I prefer not to introduce new av_ functions

xuguangxin: thanks for the explain.
could we define static functions to reduce duplicate code in one file?

xhaihao: No.  It will result in the 'duplicate definition' error when both libavxxx libraries are linked statically. In the past, we experienced such error and had to define functions with different names.

xhaihao:
this patchset was sent to the mailing list http://ffmpeg.org/pipermail/ffmpeg-devel/2021-April/278808.html

PR #384, opened 2021-04-09
fftools/ffmpeg_qsv:Change width and height of hw_frame_ctx to origina…

wenbinc-Bin:
…l value

We use w/h from vaapi hw_frame_ctx to create libva surfaces and we use Param
to init mfx-codec or mfx-vpp. All this value are already aligned value,
so we needn't to aligned width and height in qsv hw_frame_ctx.
We use hw_frame_ctx to create AVFrame, we use hw_frame_ctx->w/h to
configure crop_w/h in hwcontext_qsv.c/qsv_init_surface() function. And
this value should be original width and height, so I change the qsv's
hw_frame_ctx->width/height to original value because of these two reasons.

Signed-off-by: Wenbin Chen wenbin.chen@intel.com

### DIFF
--- a/fftools/ffmpeg_qsv.c
+++ b/fftools/ffmpeg_qsv.c
@@ -90,11 +90,13 @@ int qsv_init(AVCodecContext *s)
     frames_ctx   = (AVHWFramesContext*)ist->hw_frames_ctx->data;
     frames_hwctx = frames_ctx->hwctx;
 
-    frames_ctx->width             = FFALIGN(s->coded_width,  32);
-    frames_ctx->height            = FFALIGN(s->coded_height, 32);
+    frames_ctx->width             = s->width;
+    frames_ctx->height            = s->height;
     frames_ctx->format            = AV_PIX_FMT_QSV;
     frames_ctx->sw_format         = s->sw_pix_fmt;
     frames_ctx->initial_pool_size = 64 + s->extra_hw_frames;
+    frames_hwctx->aligned_width   = FFALIGN(s->coded_width,  32);
+    frames_hwctx->aligned_height  = FFALIGN(s->coded_height, 32);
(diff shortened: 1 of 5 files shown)


wenbinc-Bin:
Hi @xuguangxin, is this what you said that we should change width and height inside hw_frame_ctx?
There is an error about psnr and I think it is caused by aligning a large number for h264 surface. I will find a way to fix it, but I want to make sure that I am on the right direction.

xuguangxin:
Could you explain why we need align to 64?
What's happened if encode surface is mapped from 3D and it's not 64 aligned.

thanks

wenbinc-Bin:
Hi @xuguangxin :
I tested on different codec and found 64 is the largest aligned number used by MSDK so far, so I align w/h to 64. However I found this affect psnr, and I am trying to use another way to do this.

I think if the mapped surfaces is not same as hw_frame_ctx's configuration, the "hwdownload" section may have potential problem. If the mapped surface is sent directly to encoder, I think it will be ok if the aligned width and height is greater than encoder's need.

xuguangxin:
> Hi @xuguangxin :
> I tested on different codec and found 64 is the largest aligned number used by MSDK so far, so I align w/h to 64. However I found this affect psnr, and I am trying to use another way to do this.
> 
> I think if the mapped surfaces is not same as hw_frame_ctx's configuration, the "hwdownload" section may have potential problem. If the mapped surface is sent directly to encoder, I think it will be ok if the aligned width and height is greater than encoder's need.

we'd better find some MediaSDK document for 64, else it's hard to upstream the patch.
Upstream reviewer may ask, why it's not 32? If we said we tested it, reviewer may say, what if msdk changed it in some platforms?

wenbinc-Bin:
Hi @xuguangxin. I made changes on the way to aligned w/h. I keep the previous aligned value so we needn't to explain why we use 64.

PR #385, opened 2021-04-15
Remove redundant decode header

wenbinc-Bin:


wenbinc-Bin:
@xhaihao @xuguangxin @feiwan1 
could you please help me to review this patches.
Thanks

wenbinc-Bin:
@xuguangxin @feiwan1 @xhaihao 
Can anyone help me to review these patches?
Thanks

libavcodec/qsvdec.c
```diff
@@ -481,6 +481,13 @@ static int qsv_decode(AVCodecContext *avctx, QSVContext *q,
 
     } while (ret == MFX_WRN_DEVICE_BUSY || ret == MFX_ERR_MORE_SURFACE);
 
+    if (ret == MFX_ERR_INCOMPATIBLE_VIDEO_PARAM) {
+        q->reinit_flag = 1;
```
xuguangxin, line 484: If reinit_flag is true, do we need send null packet to qsv_process_data to flush the decoder?

wenbinc-Bin: Yes, it will flush decoder before re-init new decoder. you can see the code around line 647 (after "AVPacket zero_pkt = {0};")

libavcodec/qsvdec.c
```diff
@@ -637,11 +637,7 @@ static int qsv_process_data(AVCodecContext *avctx, QSVContext *q,
     if (!avctx->coded_height)
         avctx->coded_height = 720;
 
-    ret = qsv_decode_header(avctx, q, pkt, pix_fmt, &param);
-
-    if (q->reinit_flag || (ret >= 0 && (q->orig_pix_fmt != ff_qsv_map_fourcc(param.mfx.FrameInfo.FourCC) ||
-        avctx->coded_width  != param.mfx.FrameInfo.Width ||
-        avctx->coded_height != param.mfx.FrameInfo.Height))) {
+    if (q->reinit_flag || !q->session) {
```
xuguangxin: the code may misleading here.
how about we change it to 
```
if (q->reinit_flag) {
    flush...
}
if (q->reinit_flag || !q->session) {

 ret = decode_header();
 //check ret
 q->reinit_flag = 0;

}
```

xuguangxin (commented):
also, could we remove buffered_count too?
after flush, we can get buffer until drain, then set break the
```
if (q->reinit_flag) {
    decode_frame();
    if (got_output)
      return output;
}

```

wenbinc-Bin:
Got it. I will work on it.

libavcodec/qsvdec.c
```diff
@@ -497,9 +503,7 @@ static int qsv_decode(AVCodecContext *avctx, QSVContext *q,
         ++q->zero_consume_run;
         if (q->zero_consume_run > 1)
             ff_qsv_print_warning(avctx, ret, "A decode call did not consume any data");
-    } else if (!*sync && bs.DataOffset) {
-        ++q->buffered_count;
-    } else {
+    } else if (*sync) {
```
xuguangxin: why we need this if?

wenbinc-Bin: In previous code, we enter this branch when sync is assigned by some value, but when I remove the second branch the logic will change so I add this judgement to keep the logic same.

wenbinc-Bin: Well... It doesn't effect function, so removing it maybe better.

xuguangxin (approved):

xuguangxin:
could you do full round test before we send to community?
This has big impact.

wenbinc-Bin:
Ok, I will find a way to do full round test before I send it to community.

PR #399, opened 2021-05-06
libavcodec/qsvenc: fix system memory input error.

wenbinc-Bin:
vp9_qsv encoder check the UV and V value in surface when encode
frame in system memory. Assign UV+1 to V just as what sample_encode
does.

Fix command line: ffmpeg -i inputvideo -c:v vp9_qsv output.ivf

### DIFF
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -1357,6 +1357,10 @@ static int submit_frame(QSVEncContext *q, const AVFrame *frame,
         qf->surface.Data.PitchLow  = qf->frame->linesize[0];
         qf->surface.Data.Y         = qf->frame->data[0];
         qf->surface.Data.UV        = qf->frame->data[1];
+        if (frame->format == AV_PIX_FMT_NV12)
+            qf->surface.Data.V         = qf->surface.Data.UV + 1;
+        else if (frame->format == AV_PIX_FMT_P010)
+            qf->surface.Data.V         = qf->surface.Data.UV + 2;
     }
 
     qf->surface.Data.TimeStamp = av_rescale_q(frame->pts, q->avctx->time_base, (AVRational){1, 90000});


libavcodec/qsvenc.c
```diff
@@ -1357,6 +1357,7 @@ static int submit_frame(QSVEncContext *q, const AVFrame *frame,
         qf->surface.Data.PitchLow  = qf->frame->linesize[0];
         qf->surface.Data.Y         = qf->frame->data[0];
         qf->surface.Data.UV        = qf->frame->data[1];
+        qf->surface.Data.V         = qf->surface.Data.UV + 1;
```
xuguangxin: How about P010?
Is this a MediaSDK issue? for UV combined case, we only need to provide Data.UV.

wenbinc-Bin: Forget P010.
I don't know if it is a MediaSDK issue. VP9 encoder will check this two values. https://github.com/Intel-Media-SDK/MediaSDK/blob/19885ea49ceb38e0145917634e45b11c524364ec/_studio/mfx_lib/encode_hw/vp9/src/mfx_vp9_encode_hw_par.cpp#L1921

wenbinc-Bin: Add P010. I think it is not an issue, you can see how sample_encode configure surface. 
https://github.com/Intel-Media-SDK/MediaSDK/blob/19885ea49ceb38e0145917634e45b11c524364ec/samples/sample_common/src/sysmem_allocator.cpp#L105

xuguangxin: I think this is mediasdk issue. for 3 planes input it can check Y, U, V.
For 2 planes yuv, it can only check Y and UV.

wenbinc-Bin: Yes it is wired, only vp9 does this check. 
Do I need to submit issue to MSDK team? I don't think them will fix this issue.

feiwan1: It is better to fix in MSDK. Otherwise we need to add all supported formats here like 422/444 8bit/10bit.

xuguangxin: yes, we can submit a pr to mediasdk.

wenbinc-Bin: Ok, I will submit pr to mediasdk.

PR #400, opened 2021-05-08
libavfilter/dnn: Specificity in Error Handling

shubhanshu02:
## Description

Switch to returning specific error codes or DNN_GENERIC_ERROR when an error is encountered.


shubhanshu02:
@guoyejun, I have pulled the latest commits into this PR. Please have a look at the changes.

**Update**: I noticed that I missed the `DNN_ERROR` occurrences in the OpenVino backend. And new changes may also have some `DNN_ERROR` occurrences in them, so I think I should wait till all changes are upstreamed and then continue this work. Thank you.

PR #407, opened 2021-05-20
Async Support for TensorFlow Backend

shubhanshu02:
### Patch Set Description

This patchset is a part of necessary deliverables in the GSoC project [Async Support for TensorFlow Backend in FFmpeg](https://summerofcode.withgoogle.com/projects/#5224576843251712).

**Objective**: Asynchronous Support for TensorFlow backend
**Parts under this deliverable**:
    - Switch the execution mode to `TFRequestItem` based inference.
    - Implementing a standard asynchronous inference module `DNNAsyncExecModule` for use across the TF and Native backends.
    - Implement async mode in the TensorFlow backend.

### Earlier Merged Patches in this patchset

The below patches move the `TaskItem` and `InferenceItem` from the OpenVINO backend to the `dnn_backend_common` and adjust them for shared use across the three backends. Then we define the `TFRequestItem` with execution parameters and switch the execution mechanism in the TensorFlow backend to `TFRequestItem` based inference.

f5ab890 lavfi/dnn: Extract TaskItem and InferenceItem from OpenVino Backend
446b4f7 lavfi/dnn: Convert output_name to char** in TaskItem
9675ebb lavfi/dnn: Add nb_output to TaskItem
6b961f7 lavfi/dnn: Use uint8_t for async and do_ioproc in TaskItems
5509235 lavfi/dnn: Fill Task using Common Function
68cf14d lavfi/dnn_backend_tf: TaskItem Based Inference
a4de605 lavfi/dnn_backend_tf: Add TFInferRequest and TFRequestItem
08d8b3b lavfi/dnn_backend_tf: Request-based Execution
b849228 lavfi/dnn_backend_tf: Separate function for filling RequestItem
84e4e60 lavfi/dnn_backend_tf: Separate function for Completion Callback
6f9570a lavfi/dnn_backend_tf: Error Handling

### Final Patches

The below-mentioned patches implement the `DNNAsyncExecModule` and use them in the TensorFlow backend for adding the async mode. The methodology behind the `DNNAsyncExecModule` being to execute a number of `TFRequestItem`'s (which can also be set using the backend configuration parameter `nireq`) concurrently along the main FFmpeg execution thread so that the inference requests can be executed in an asynchronous fashion.

Each `TFRequestItem` has its instance of `DNNAsyncExecModule`, which corresponds to a single thread. When `TF_SessionRun` returns, the thread routine also returns with relevant exit code, and the `TFRequestItem` is pushed back to the `request_queue`. This return status is caught when next time the same `TFRequestItem` is used for the execution. In case the previous execution failed, the error message is already printed, and we cancel all further executions by returning `DNN_ERROR`.

86f0a4f9 lavfi/dnn: Add Async Execution Mechanism and Documentation
c7165785 lavfi/dnn: Common Function to Get Async Result in DNN Backends
e6ae8fc1 lavfi/dnn_backend_tf: TFInferRequest Execution and Documentation
0985e928 lavfi/dnn: Async Support for TensorFlow Backend
a3db9b54 lavfi/dnn_backend_tf: Error Handling for execute_model_tf
4d627ace lavfi/dnn_backend_tf: Add TF_Status to TFRequestItem
009b2e5b lavfi/dnn: Extract Common Parts from get_output functions
371e5672 lavfi/dnn_backend_tf: Error Handling for tf_create_inference_request
2063745a lavfi/dnn: DNNAsyncExecModule Execution Failure Handling

### DIFF
--- a/libavfilter/dnn/dnn_backend_common.c
+++ b/libavfilter/dnn/dnn_backend_common.c
@@ -69,3 +69,56 @@ DNNReturnType ff_dnn_fill_task(TaskItem *task, DNNExecBaseParams *exec_params, v
 
     return DNN_SUCCESS;
 }
+
+/**
+ * Thread routine for async execution.
+ * @param args pointer to DNNAsyncExecModule module
+ */
+static void *async_thread_routine(void *args)
+{
+    DNNAsyncExecModule *async_module = args;
+    void *request = async_module->args;
+
+    async_module->start_inference(request);
+    async_module->callback(request);
+    return NULL;
+}
+
+DNNReturnType ff_dnn_async_module_cleanup(DNNAsyncExecModule *async_module)
+{
+    if (!async_module) {
+        return DNN_ERROR;
+    }
+#if HAVE_PTHREAD_CANCEL
+    pthread_join(async_module->thread_id, NULL);
+#endif
+    async_module->start_inference = NULL;
+    async_module->callback = NULL;
+    async_module->args = NULL;
+    return DNN_SUCCESS;
+}
+
+DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_module)
+{
+    int ret;
+
+    if (!async_module) {
+        av_log(ctx, AV_LOG_ERROR, "async_module is null when starting async inference.\n");
+        return DNN_ERROR;
+    }
+
+#if HAVE_PTHREAD_CANCEL
+    pthread_join(async_module->thread_id, NULL);
+    ret = pthread_create(&async_module->thread_id, NULL, async_thread_routine, async_module);
+    if (ret != 0) {
+        av_log(ctx, AV_LOG_ERROR, "Unable to start async inference.\n");
+        return DNN_ERROR;
+    }
+#else
+    if (async_module->start_inference(async_module->args) != DNN_SUCCESS) {
+        return DNN_ERROR;
+    }
+    async_module->callback(async_module->args);
+#endif
+    return DNN_SUCCESS;
+}
--- a/libavfilter/dnn/dnn_backend_common.c
+++ b/libavfilter/dnn/dnn_backend_common.c
@@ -23,6 +23,9 @@
 
 #include "dnn_backend_common.h"
 
+#define DNN_ASYNC_SUCCESS (void *)0
+#define DNN_ASYNC_FAIL (void *)-1
+
 int ff_check_exec_params(void *ctx, DNNBackendType backend, DNNFunctionType func_type, DNNExecBaseParams *exec_params)
 {
     if (!exec_params) {
@@ -79,18 +82,25 @@ static void *async_thread_routine(void *args)
     DNNAsyncExecModule *async_module = args;
     void *request = async_module->args;
 
-    async_module->start_inference(request);
+    if (async_module->start_inference(request) != DNN_SUCCESS) {
+        return DNN_ASYNC_FAIL;
+    }
     async_module->callback(request);
-    return NULL;
+    return DNN_ASYNC_SUCCESS;
 }
 
 DNNReturnType ff_dnn_async_module_cleanup(DNNAsyncExecModule *async_module)
 {
+    void *status = 0;
     if (!async_module) {
         return DNN_ERROR;
     }
 #if HAVE_PTHREAD_CANCEL
-    pthread_join(async_module->thread_id, NULL);
+    pthread_join(async_module->thread_id, &status);
+    if (status == DNN_ASYNC_FAIL) {
+        av_log(NULL, AV_LOG_ERROR, "Last Inference Failed.\n");
+        return DNN_ERROR;
+    }
 #endif
     async_module->start_inference = NULL;
     async_module->callback = NULL;
@@ -101,6 +111,7 @@ DNNReturnType ff_dnn_async_module_cleanup(DNNAsyncExecModule *async_module)
 DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_module)
 {
     int ret;
+    void *status = 0;
 
     if (!async_module) {
         av_log(ctx, AV_LOG_ERROR, "async_module is null when starting async inference.\n");
@@ -108,7 +119,11 @@ DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_
     }
 
 #if HAVE_PTHREAD_CANCEL
-    pthread_join(async_module->thread_id, NULL);
+    pthread_join(async_module->thread_id, &status);
+    if (status == DNN_ASYNC_FAIL) {
+        av_log(ctx, AV_LOG_ERROR, "Unable to start inference as previous inference failed.\n");
+        return DNN_ERROR;
+    }
     ret = pthread_create(&async_module->thread_id, NULL, async_thread_routine, async_module);
     if (ret != 0) {
         av_log(ctx, AV_LOG_ERROR, "Unable to start async inference.\n");
--- a/libavfilter/dnn/dnn_backend_common.c
+++ b/libavfilter/dnn/dnn_backend_common.c
@@ -122,3 +122,23 @@ DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_
 #endif
     return DNN_SUCCESS;
 }
+
+DNNAsyncStatusType ff_dnn_get_async_result_common(Queue *task_queue, AVFrame **in, AVFrame **out)
+{
+    TaskItem *task = ff_queue_peek_front(task_queue);
+
+    if (!task) {
+        return DAST_EMPTY_QUEUE;
+    }
+
+    if (task->inference_done != task->inference_todo) {
+        return DAST_NOT_READY;
+    }
+
+    *in = task->in_frame;
+    *out = task->out_frame;
+    ff_queue_pop_front(task_queue);
+    av_freep(&task);
+
+    return DAST_SUCCESS;
+}
--- a/libavfilter/dnn/dnn_backend_common.c
+++ b/libavfilter/dnn/dnn_backend_common.c
@@ -142,3 +142,29 @@ DNNAsyncStatusType ff_dnn_get_async_result_common(Queue *task_queue, AVFrame **i
 
     return DAST_SUCCESS;
 }
+
+DNNReturnType ff_dnn_fill_gettingoutput_task(TaskItem *task, DNNExecBaseParams *exec_params, void *backend_model, int input_height, int input_width, void *ctx)
+{
+    AVFrame *in_frame = NULL;
+    AVFrame *out_frame = NULL;
+
+    in_frame = av_frame_alloc();
+    if (!in_frame) {
+        av_log(ctx, AV_LOG_ERROR, "Failed to allocate memory for input frame\n");
+        return DNN_ERROR;
+    }
+
+    out_frame = av_frame_alloc();
+    if (!out_frame) {
+        av_frame_free(&in_frame);
+        av_log(ctx, AV_LOG_ERROR, "Failed to allocate memory for output frame\n");
+        return DNN_ERROR;
+    }
+
+    in_frame->width = input_width;
+    in_frame->height = input_height;
+    exec_params->in_frame = in_frame;
+    exec_params->out_frame = out_frame;
+
+    return ff_dnn_fill_task(task, exec_params, backend_model, 0, 0);
+}
--- a/libavfilter/dnn/dnn_backend_common.h
+++ b/libavfilter/dnn/dnn_backend_common.h
@@ -24,6 +24,7 @@
 #ifndef AVFILTER_DNN_DNN_BACKEND_COMMON_H
 #define AVFILTER_DNN_DNN_BACKEND_COMMON_H
 
+#include "queue.h"
 #include "../dnn_interface.h"
 #include "libavutil/thread.h"
 
@@ -121,5 +122,19 @@ DNNReturnType ff_dnn_async_module_cleanup(DNNAsyncExecModule *async_module);
  * @retval DNN_ERROR in case async inference cannot be started
  */
 DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_module);
+
+/**
+ * Extract input and output frame from the Task Queue after
+ * asynchronous inference.
+ *
+ * @param task_queue pointer to the task queue of the backend
+ * @param in double pointer to the input frame
+ * @param out double pointer to the output frame
+ *
+ * @retval DAST_EMPTY_QUEUE if task queue is empty
+ * @retval DAST_NOT_READY if inference not completed yet.
+ * @retval DAST_SUCCESS if result successfully extracted
+ */
+DNNAsyncStatusType ff_dnn_get_async_result_common(Queue *task_queue, AVFrame **in, AVFrame **out);
 
 #endif
--- a/libavfilter/dnn/dnn_backend_common.h
+++ b/libavfilter/dnn/dnn_backend_common.h
@@ -25,6 +25,7 @@
 #define AVFILTER_DNN_DNN_BACKEND_COMMON_H
 
 #include "../dnn_interface.h"
+#include "libavutil/thread.h"
 
 #define DNN_BACKEND_COMMON_OPTIONS \
     { "nireq",           "number of request",             OFFSET(options.nireq),           AV_OPT_TYPE_INT,    { .i64 = 0 },     0, INT_MAX, FLAGS },
@@ -49,6 +50,34 @@ typedef struct InferenceItem {
     uint32_t bbox_index;
 } InferenceItem;
 
+/**
+ * Common Async Execution Mechanism for the DNN Backends.
+ */
+typedef struct DNNAsyncExecModule {
+    /**
+     * Synchronous inference function for the backend
+     * with corresponding request item as the argument.
+     */
+    DNNReturnType (*start_inference)(void *request);
+
+    /**
+     * Completion Callback for the backend.
+     * Expected argument type of callback must match that
+     * of the inference function.
+     */
+    void (*callback)(void *args);
+
+    /**
+     * Argument for the execution functions.
+     * i.e. Request item for the backend.
+     */
+    void *args;
+#if HAVE_PTHREAD_CANCEL
+    pthread_t thread_id;
+    pthread_attr_t thread_attr;
+#endif
+} DNNAsyncExecModule;
+
 int ff_check_exec_params(void *ctx, DNNBackendType backend, DNNFunctionType func_type, DNNExecBaseParams *exec_params);
 
 /**
@@ -65,5 +94,32 @@ int ff_check_exec_params(void *ctx, DNNBackendType backend, DNNFunctionType func
  * @retval DNN_ERROR if flags are invalid or any parameter is NULL
  */
 DNNReturnType ff_dnn_fill_task(TaskItem *task, DNNExecBaseParams *exec_params, void *backend_model, int async, int do_ioproc);
+
+/**
+ * Join the Async Execution thread and set module pointers to NULL.
+ *
+ * @param async_module pointer to DNNAsyncExecModule module
+ *
+ * @retval DNN_SUCCESS if successful
+ * @retval DNN_ERROR if async_module is NULL
+ */
+DNNReturnType ff_dnn_async_module_cleanup(DNNAsyncExecModule *async_module);
+
+/**
+ * Start asynchronous inference routine for the TensorFlow
+ * model on a detached thread. It calls the completion callback
+ * after the inference completes. Completion callback and inference
+ * function must be set before calling this function.
+ *
+ * If POSIX threads aren't supported, the execution rolls back
+ * to synchronous mode, calling completion callback after inference.
+ *
+ * @param ctx pointer to the backend context
+ * @param async_module pointer to DNNAsyncExecModule module
+ *
+ * @retval DNN_SUCCESS on the start of async inference.
+ * @retval DNN_ERROR in case async inference cannot be started
+ */
+DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_module);
 
 #endif
--- a/libavfilter/dnn/dnn_backend_common.h
+++ b/libavfilter/dnn/dnn_backend_common.h
@@ -137,4 +137,20 @@ DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_
  */
 DNNAsyncStatusType ff_dnn_get_async_result_common(Queue *task_queue, AVFrame **in, AVFrame **out);
 
+/**
+ * Allocate input and output frames and fill the Task
+ * with execution parameters.
+ *
+ * @param task pointer to the allocated task
+ * @param exec_params pointer to execution parameters
+ * @param backend_model void pointer to the backend model
+ * @param input_height height of input frame
+ * @param input_width width of input frame
+ * @param ctx pointer to the backend context
+ *
+ * @retval DNN_SUCCESS if successful
+ * @retval DNN_ERROR if allocation fails
+ */
+DNNReturnType ff_dnn_fill_gettingoutput_task(TaskItem *task, DNNExecBaseParams *exec_params, void *backend_model, int input_height, int input_width, void *ctx);
+
 #endif
--- a/libavfilter/dnn/dnn_backend_openvino.c
+++ b/libavfilter/dnn/dnn_backend_openvino.c
@@ -640,10 +640,15 @@ static DNNReturnType get_output_ov(void *model, const char *input_name, int inpu
     OVContext *ctx = &ov_model->ctx;
     TaskItem task;
     OVRequestItem *request;
-    AVFrame *in_frame = NULL;
-    AVFrame *out_frame = NULL;
     IEStatusCode status;
     input_shapes_t input_shapes;
+    DNNExecBaseParams exec_params = {
+        .input_name     = input_name,
+        .output_names   = &output_name,
+        .nb_output      = 1,
+        .in_frame       = NULL,
+        .out_frame      = NULL,
+    };
 
     if (ov_model->model->func_type != DFT_PROCESS_FRAME) {
         av_log(ctx, AV_LOG_ERROR, "Get output dim only when processing frame.\n");
@@ -669,51 +674,29 @@ static DNNReturnType get_output_ov(void *model, const char *input_name, int inpu
         }
     }
 
-    in_frame = av_frame_alloc();
-    if (!in_frame) {
-        av_log(ctx, AV_LOG_ERROR, "Failed to allocate memory for input frame\n");
+    if (ff_dnn_fill_gettingoutput_task(&task, &exec_params, ov_model, input_height, input_width, ctx) != DNN_SUCCESS) {
         return DNN_ERROR;
     }
-    in_frame->width = input_width;
-    in_frame->height = input_height;
-
-    out_frame = av_frame_alloc();
-    if (!out_frame) {
-        av_log(ctx, AV_LOG_ERROR, "Failed to allocate memory for output frame\n");
-        av_frame_free(&in_frame);
-        return DNN_ERROR;
-    }
-
-    task.do_ioproc = 0;
-    task.async = 0;
-    task.input_name = input_name;
-    task.in_frame = in_frame;
-    task.output_names = &output_name;
-    task.out_frame = out_frame;
-    task.nb_output = 1;
-    task.model = ov_model;
 
     if (extract_inference_from_task(ov_model->model->func_type, &task, ov_model->inference_queue, NULL) != DNN_SUCCESS) {
-        av_frame_free(&out_frame);
-        av_frame_free(&in_frame);
         av_log(ctx, AV_LOG_ERROR, "unable to extract inference from task.\n");
-        return DNN_ERROR;
+        ret = DNN_ERROR;
+        goto err;
     }
 
     request = ff_safe_queue_pop_front(ov_model->request_queue);
     if (!request) {
-        av_frame_free(&out_frame);
-        av_frame_free(&in_frame);
         av_log(ctx, AV_LOG_ERROR, "unable to get infer request.\n");
-        return DNN_ERROR;
+        ret = DNN_ERROR;
+        goto err;
     }
 
     ret = execute_model_ov(request, ov_model->inference_queue);
-    *output_width = out_frame->width;
-    *output_height = out_frame->height;
-
-    av_frame_free(&out_frame);
-    av_frame_free(&in_frame);
+    *output_width = task.out_frame->width;
+    *output_height = task.out_frame->height;
+err:
+    av_frame_free(&task.out_frame);
+    av_frame_free(&task.in_frame);
     return ret;
 }
 
--- a/libavfilter/dnn/dnn_backend_openvino.c
+++ b/libavfilter/dnn/dnn_backend_openvino.c
@@ -32,7 +32,6 @@
 #include "libavutil/avstring.h"
 #include "libavutil/detection_bbox.h"
 #include "../internal.h"
-#include "queue.h"
 #include "safe_queue.h"
 #include <c_api/ie_c_api.h>
 #include "dnn_backend_common.h"
@@ -883,22 +882,7 @@ DNNReturnType ff_dnn_execute_model_async_ov(const DNNModel *model, DNNExecBasePa
 DNNAsyncStatusType ff_dnn_get_async_result_ov(const DNNModel *model, AVFrame **in, AVFrame **out)
 {
     OVModel *ov_model = model->model;
-    TaskItem *task = ff_queue_peek_front(ov_model->task_queue);
-
-    if (!task) {
-        return DAST_EMPTY_QUEUE;
-    }
-
-    if (task->inference_done != task->inference_todo) {
-        return DAST_NOT_READY;
-    }
-
-    *in = task->in_frame;
-    *out = task->out_frame;
-    ff_queue_pop_front(ov_model->task_queue);
-    av_freep(&task);
-
-    return DAST_SUCCESS;
+    return ff_dnn_get_async_result_common(ov_model->task_queue, in, out);
 }
 
 DNNReturnType ff_dnn_flush_ov(const DNNModel *model)
--- a/libavfilter/dnn/dnn_backend_tf.c
+++ b/libavfilter/dnn/dnn_backend_tf.c
@@ -95,6 +95,13 @@ static void free_buffer(void *data, size_t length)
     av_freep(&data);
 }
 
+/**
+ * Free the contents of TensorFlow inference request.
+ * It does not free the TFInferRequest instance.
+ *
+ * @param request pointer to TFInferRequest instance.
+ * NULL pointer is allowed.
+ */
 static void tf_free_request(TFInferRequest *request)
 {
     if (!request)
@@ -117,6 +124,12 @@ static void tf_free_request(TFInferRequest *request)
     }
 }
 
+/**
+ * Create a TensorFlow inference request. All properties
+ * are initially unallocated and set as NULL.
+ *
+ * @return pointer to the allocated TFInferRequest instance.
+ */
 static TFInferRequest *tf_create_inference_request(void)
 {
     TFInferRequest *infer_request = av_malloc(sizeof(TFInferRequest));
@@ -125,6 +138,38 @@ static TFInferRequest *tf_create_inference_request(void)
     infer_request->input_tensor = NULL;
     infer_request->output_tensors = NULL;
     return infer_request;
+}
+
+/**
+ * Start synchronous inference for the TensorFlow model.
+ *
+ * @param request pointer to the TFRequestItem for inference
+ * @retval DNN_SUCCESS if execution is successful
+ * @retval DNN_ERROR if execution fails
+ */
+static DNNReturnType tf_start_inference(void *args)
+{
+    TFRequestItem *request = args;
+    TFInferRequest *infer_request = request->infer_request;
+    InferenceItem *inference = request->inference;
+    TaskItem *task = inference->task;
+    TFModel *tf_model = task->model;
+
+    if (!request) {
+        av_log(&tf_model->ctx, AV_LOG_ERROR, "TFRequestItem is NULL\n");
+        return DNN_ERROR;
+    }
+
+    TF_SessionRun(tf_model->session, NULL,
+                  infer_request->tf_input, &infer_request->input_tensor, 1,
+                  infer_request->tf_outputs, infer_request->output_tensors,
+                  task->nb_output, NULL, 0, NULL,
+                  tf_model->status);
+    if (TF_GetCode(tf_model->status) != TF_OK) {
+        av_log(&tf_model->ctx, AV_LOG_ERROR, "%s", TF_Message(tf_model->status));
+        return DNN_ERROR;
+    }
+    return DNN_SUCCESS;
 }
 
 static DNNReturnType extract_inference_from_task(TaskItem *task, Queue *inference_queue)
--- a/libavfilter/dnn/dnn_backend_tf.c
+++ b/libavfilter/dnn/dnn_backend_tf.c
@@ -75,6 +75,7 @@ typedef struct TFInferRequest {
 typedef struct TFRequestItem {
     TFInferRequest *infer_request;
     InferenceItem *inference;
+    TF_Status *status;
     DNNAsyncExecModule exec_module;
 } TFRequestItem;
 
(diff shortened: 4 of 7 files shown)


libavfilter/dnn/dnn_backend_common.h
```diff
@@ -26,6 +26,42 @@
...
 
+// one task for one function call from dnn interface
+typedef struct TaskItem {
+    void *model;
+    AVFrame *in_frame;
+    AVFrame *out_frame;
+    const char *input_name;
+    const char **output_names;
+    int async;
+    int do_ioproc;
```
shubhanshu02: Since these have value 0 or 1, will `uint8_t` be better?

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -187,7 +241,27 @@ static DNNReturnType get_output_tf(void *model, const char *input_name, int inpu
...
 
-    ret = execute_model_tf(tf_model->model, input_name, in_frame, &output_name, 1, out_frame, 0);
+    task.do_ioproc = 0;
+    task.async = 0;
+    task.input_name = input_name;
+    task.in_frame = in_frame;
+    task.output_names = &output_name;
+    task.out_frame = out_frame;
+    task.model = tf_model;
+    task.nb_output = 1;
```
shubhanshu02: This part and the below part will remain common in all three backends. Shall we use a common function for this?
```C
    if (!in_frame) {
        av_log(ctx, AV_LOG_ERROR, "Failed to allocate memory for input frame\n");
        return DNN_ERROR;
    }
    out_frame = av_frame_alloc();
    if (!out_frame) {
        av_log(ctx, AV_LOG_ERROR, "Failed to allocate memory for output frame\n");
        av_frame_free(&amp;in_frame);
        return DNN_ERROR;
    }
    in_frame->width = input_width;
    in_frame->height = input_height;
```

shubhanshu02 (commented):
@guoyejun I have these two questions regarding these changes.

guoyejun:
> This PR proposes the following changes:
> 
> 1. Extract `TaskItem` and `InferenceItem` from the OpenVino backend.
> 2. Change `output_name` to `output_names` in `TaskItem` (for other backends).
> 3. Common function for filling tasks in `ff_dnn_execute_model_<backend>` functions
> 4. In `ff_dnn_free_model_ov`, objects popped from `inference_queue` must have type `* InferenceItem`
> 5. RequestItem based execution in TensorFlow backend.
> 
> ### Description
> * Each RequestItem contains a pointer to a `tf_infer_request` instance. The `infer_request` contains the parameters we need to store the input and output of the `TF_SessionRun`.
> * Initially, when the model is loaded in `ff_dnn_load_model_tf`, the request queue is created, and a total of `nireq` requests are pushed to it. At this step, we also allocate the space for the `tf_infer_request` for each request using the function `tf_create_inference_request`.
> * After completion of `TF_SessionRun`, `infer_completion_callback` is called, and the output frame is fetched. Before pushing the `RequestItem` back into the Safe Queue, the pointers in `tf_infer_request` are freed, keeping itself allocated (so we don't need to allocate its space again and again).

it doesn't matter if you leave this blank since there's commit log in each change. 
btw, I have something others to do recently, so my review will be a bit slow.

libavfilter/dnn/dnn_backend_common.h
guoyejun: yes

libavfilter/dnn/dnn_backend_tf.c
guoyejun: you can try it with a small separate patch

libavfilter/dnn/dnn_backend_common.h
```diff
@@ -26,6 +26,26 @@
 
 #include "../dnn_interface.h"
 
+// one task for one function call from dnn interface
+typedef struct TaskItem {
+    void *model;
```
guoyejun: add a simple comment to explain what model is

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -184,7 +165,7 @@ static DNNReturnType fill_model_input_ov(OVModel *ov_model, RequestItem *request
         request->inferences[i] = inference;
         request->inference_count = i + 1;
         task = inference->task;
-        switch (task->ov_model->model->func_type) {
+        switch (ov_model->model->func_type) {
```
guoyejun: you may extract it into a new separate small patch, so I can have an easy life for review.  I mean all the changes 
from task->ov_model to ov_model.

shubhanshu02:
> btw, I have something others to do recently, so my review will be a bit slow.

Sure, no problem.

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Okay

libavfilter/dnn/dnn_backend_common.h
shubhanshu02: Sure

libavfilter/dnn/dnn_backend_openvino.c
shubhanshu02: Sure. I divide the changes in the OpenVino backend into separate smaller patches.

Currently this patch had the following changes:
1. Extract TaskItems and InferenceItems to `dnn_backend_common.h`
2. Change model in TaskItem to `void *model` and the task->ov_model to ov_model part. (since `task->ov_model` will be a void pointer, so its properties like `task->ov_model->model` cannot be accessed directly without casting it to `((OVModel *) task->ov_model)->model`.
3. Change `char *output_name` to `char **output_names` and corresponding changes in `dnn_backend_openvino.c`

shubhanshu02: But I'm wondering each OVModel in a task points to the same DNNModel instance. So why use `task->ov_model->model` instead of `ov_model` (like ov_model parameter in `fill_model_input_ov`)?

guoyejun: I might forget ov_model when I wrote the code.

shubhanshu02: No problem. I'll add those changes as well.

guoyejun (commented):
typo in commit log line and body:  uint8_32 -> uint8_t

guoyejun:
> typo in commit log line and body: uint8_32 -> uint8_t

for the patch: lavfi/dnn: Use uint8_32 for async and do_ioproc in TaskItems

libavfilter/dnn/dnn_backend_common.c
```diff
@@ -49,3 +49,27 @@ int ff_check_exec_params(void *ctx, DNNBackendType backend, DNNFunctionType func
...
+    task->model = backend_model;
+    task->nb_output = exec_params->nb_output;
+
+    if (backend == DNN_OV)
+        task->output_names = &exec_params->output_names[0];
+    else
+        task->output_names = exec_params->output_names;
+
+    return DNN_SUCCESS;
+}
```
guoyejun: need a new line at the end of the file

libavfilter/dnn/dnn_backend_common.c
```diff
@@ -49,3 +49,27 @@ int ff_check_exec_params(void *ctx, DNNBackendType backend, DNNFunctionType func
...
+    task->do_ioproc = do_ioproc;
+    task->async = async;
+    task->input_name = exec_params->input_name;
+    task->in_frame = exec_params->in_frame;
+    task->out_frame = exec_params->out_frame;
+    task->model = backend_model;
+    task->nb_output = exec_params->nb_output;
+
+    if (backend == DNN_OV)
+        task->output_names = &exec_params->output_names[0];
```
guoyejun: it is the same as:
task->output_names = exec_params->output_names

so backend is not needed

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -941,7 +941,7 @@ void ff_dnn_free_model_ov(DNNModel **model)
         ff_safe_queue_destroy(ov_model->request_queue);
 
         while (ff_queue_size(ov_model->inference_queue) != 0) {
-            TaskItem *item = ff_queue_pop_front(ov_model->inference_queue);
+            InferenceItem *item = ff_queue_pop_front(ov_model->inference_queue);
```
guoyejun: can you first send out this patch to community?

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -36,6 +36,7 @@
 #include "dnn_io_proc.h"
 #include "dnn_backend_common.h"
 #include <tensorflow/c/c_api.h>
+#include <pthread.h>
```
guoyejun: #include "libavutil/thread.h"

btw, i today do not have more time to review starting from the tf backend,

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -1078,6 +1210,8 @@ void ff_dnn_free_model_tf(DNNModel **model)
         tf_model = (*model)->model;
         while (ff_safe_queue_size(tf_model->request_queue) != 0) {
             RequestItem *item = ff_safe_queue_pop_front(tf_model->request_queue);
+
```
guoyejun: extra empty line

libavfilter/dnn/dnn_backend_openvino.c
shubhanshu02: Sure

libavfilter/dnn/dnn_backend_common.c
shubhanshu02: Oh, okay. Corrected

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: I have corrected this and handled the case where pthreads aren't supported. 
No problem, you may review whenever you find the time. I'll continue further.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -731,170 +836,240 @@ DNNModel *ff_dnn_load_model_tf(const char *model_filename, DNNFunctionType func_
     model->func_type = func_type;
 
     return model;
+err:
+    ff_dnn_free_model_tf(&model);
```
guoyejun: is it possible to separate the error handle into a new small patch?

as a reviewer, hope to see small but function complete patch one by one.

btw, you may send these patches into the community just to let others know what you are doing.  I'll not review those patches, I'll continue to do the first round review here, and you can send v2 patches when we finish here.

shubhanshu02: I understand these changes in a single patch roughly amount to 600 lines, making it harder to read. I'll convert these to smaller patches.

Sure, I'll send these today.

guoyejun: it is still a big patch, possible to separate it more?  It's fine to say 'no' if there's really no reasonable method.

guoyejun:
you may send out the first 5 patches to community.  I still do not have time for the 6th patch (the first tf big patch).

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: One way I could think of is to convert the patch ("lavfi/dnn_backend_tf: Request-based Execution") to two smaller patches -

1. One where execution is changed to only TaskItem based execution (this will account for all the changes like `output_names` to `task->output_names`) - Bigger Patch as compared to below one
2. Add RequestItem related changes to the above 

I'll try these changes and update them.

shubhanshu02:
> you may send out the first 5 patches to community. I still do not have time for the 6th patch (the first tf big patch).

Sure. I was thinking should I include only the async patchset in this PR, so that it'd a bit easier to review? Since the unification part is still in progress, I will add the patches related to that once async changes get reviewed.

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Sir, I tried partitioning the changes further down to smaller patches. The async execution functions for `tf_infer_request` are also now in separate patches.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -171,26 +348,52 @@ static DNNReturnType get_output_tf(void *model, const char *input_name, int inpu
...
+        av_log(ctx, AV_LOG_ERROR, "unable to extract inference from task.\n");
+        ret = DNN_ERROR;
+        goto final;
+    }
+
+    request = ff_safe_queue_pop_front(tf_model->request_queue);
+    if (!request) {
+        av_log(ctx, AV_LOG_ERROR, "unable to get infer request.\n");
+        ret = DNN_ERROR;
+        goto final;
```
guoyejun: use goto err or something like, I do not see ffmpeg use 'final' for the label.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -52,26 +56,199 @@ typedef struct TFModel{
...
+} tf_infer_request;
+
+typedef struct RequestItem {
+    tf_infer_request *infer_request;
+    InferenceItem *inference;
+#if HAVE_PTHREAD_CANCEL
+    pthread_t thread;
+    pthread_attr_t thread_attr;
+#endif
+} RequestItem;
```
guoyejun: to make it clear, change to TFRequestItem, (and so OVRquestItem for openvino backend),

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -52,26 +56,199 @@ typedef struct TFModel{
...
+/**
+ * Stores execution parameters for single
+ * call to the TensorFlow C API
+ */
+typedef struct tf_infer_request {
+    TF_Output *tf_outputs;
+    TF_Tensor **output_tensors;
+    TF_Output *tf_input;
+    TF_Tensor *input_tensor;
+} tf_infer_request;
```
guoyejun: ffmpeg's typedef name style:  TFInferRequest

guoyejun:
did you see the performance change when async is enabled?  You may add it into commit log (don't need to mention the specific CPU model/vendor)

shubhanshu02:
> did you see the performance change when async is enabled? You may add it into commit log (don't need to mention the specific CPU model/vendor)

Yes, there was some performance gain as follows.

1. When I run the SRCNN model using the following command:

```bash
time ffmpeg -i 480p.mp4 -vf format=yuv420p,scale=w=iw*2:h=ih*2,dnn_processing=dnn_backend=tensorflow:model=srcnn.pb:input=x:output=y -y srcnn.mp4
```
the output is as follows with master branch build:
```
real	6m13.723s
user	41m55.029s
sys	2m43.433s
```
and for these commits:
```
real	5m49.242s
user	39m11.592s
sys	3m38.245s
```

2. When I run the same video with de-rain model on the current master build, the output is:
```
real	2m8.719s
user	10m21.391s
sys	0m22.401s
```
and with the async patches:
```
real	1m13.911s
user	8m36.504s
sys	0m20.498s
```

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Sure

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Okay, thanks. Corrected.

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Yes, actually, I was a bit confused about what to use for its name. Thanks.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -723,6 +754,7 @@ DNNModel *ff_dnn_load_model_tf(const char *model_filename, DNNFunctionType func_
         }
     }
 
+    tf_model->inference_queue = ff_queue_create();
```
guoyejun: if (!tf_model->inference_queue) {
free...
return NULL;
}

on other comment for the first patch.

shubhanshu02: Yes, the error handling of this part is in the commit "lavfi/dnn_backend_tf: Error Handling".

guoyejun: ok, so no problem of the first patch.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -72,6 +89,29 @@ static void free_buffer(void *data, size_t length)
...
+{
+    if (!request)
+        return;
+    if (request->input_tensor) {
+        TF_DeleteTensor(request->input_tensor);
+        request->input_tensor = NULL;
+    }
+    av_freep(&request->tf_input);
+    av_freep(&request->tf_outputs);
+    av_freep(&request->output_tensors);
```
guoyejun: the below are missed?

TF_DeleteTensor(request->output_tensors[i]);
av_freep(&request);

shubhanshu02: Oh, yes! TF_DeleteTensor for output tensors is missed. I'll add it.

Actually, this function frees only the contents of the request (used in the completion callback). So the same request item can be used again after inference. That's why the request pointer is not freed.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -991,6 +1008,14 @@ void ff_dnn_free_model_tf(DNNModel **model)
 
     if (*model){
         tf_model = (*model)->model;
+        while (ff_safe_queue_size(tf_model->request_queue) != 0) {
+            TFRequestItem *item = ff_safe_queue_pop_front(tf_model->request_queue);
+            tf_free_request(item->infer_request);
+            av_freep(&item->infer_request);
+            av_freep(&item);
+        }
+        ff_safe_queue_destroy(tf_model->request_queue);
```
guoyejun, line 1237: I did a quick review of patch: lavfi/dnn_backend_tf: Request-based Execution. I did not find obvious issues, the basic code logic is right. So now no comment for this patch.

shubhanshu02: Okay, thanks.

guoyejun: btw, I'm thinking the name of inference queue / InferenceItem.  Looks that they are not the best because of the existing of batch mode.  One inference item is not for one inference API call to the framework (TF/OpenVINO/Native), that is what requestItem/infer_request/request_queue does.  How about to change InferenceItem to SubtaskItem?  Do you have a better naming?  @Semmer2  @shubhanshu02  

The rename can be done after all of these patches in this patch set are accepted in upstream.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -906,71 +902,98 @@ static DNNReturnType execute_model_tf(TFRequestItem *request, Queue *inference_q
...
+        av_freep(&outputs);
+        ff_safe_queue_push_back(tf_model->request_queue, request);
+        return (task->inference_done == task->inference_todo) ? DNN_SUCCESS : DNN_ERROR;
     }
-    task->inference_done++;
-    tf_free_request(infer_request);
-    av_freep(&outputs);
-    ff_safe_queue_push_back(tf_model->request_queue, request);
-    return (task->inference_done == task->inference_todo) ? DNN_SUCCESS : DNN_ERROR;
 }
```
guoyejun, line 1120: basically ok for patch: lavfi/dnn_backend_tf: Separate function for filling RequestItem.

guoyejun: basically ok for patch: lavfi/dnn_backend_tf: Separate function for Completion Callback

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Yes, the name `InferenceItem` is a bit misleading. The name `SubTaskItem`, as you suggested, seems good for naming since each `InferenceItem` is a part of a `TaskItem`.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -983,6 +1025,10 @@ static DNNReturnType execute_model_tf(TFRequestItem *request, Queue *inference_q
     TaskItem *task;
 
     inference = ff_queue_peek_front(inference_queue);
+    if (!inference) {
+        av_log(NULL, AV_LOG_ERROR, "Failed to get inference item\n");
+        return DNN_ERROR;
+    }
     task = inference->task;
```
guoyejun, line 1094: basically ok for patch: lavfi/dnn_backend_tf: Error Handling.

You may send the first 6 patches to mail list. 

It is async functions starting from the next patch, and I'll find other time for the review.

shubhanshu02: Okay. I'll send them once I add the deletion of output_tensors in `tf_free_request`.

Thank you for reviewing.

guoyejun: well done, thanks.

guoyejun:
I've just pushed the first 6 patches, you may send patches such as 33d7157 and a8c07a4, which are not relative to tensorflow async, to the community.

shubhanshu02:
> I've just pushed the first 6 patches, you may send patches such as [33d7157](https://github.com/intel-media-ci/ffmpeg/commit/33d7157dec1d86e1c12c339741eabf689e284fc2) and [a8c07a4](https://github.com/intel-media-ci/ffmpeg/commit/a8c07a4fab3ead37a3aa0a6be825d0f82f7aa1ca), which are not relative to tensorflow async, to the community.

Sure, I'll send them soon. Thank you.

libavfilter/dnn/queue.h
```diff
@@ -18,6 +18,7 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
  */
 
+#include <stddef.h>
```
guoyejun, line 21: why we need this line?

libavfilter/dnn/dnn_backend_common.h
```diff
@@ -80,5 +81,6 @@ DNNReturnType ff_dnn_fill_task(TaskItem *task, DNNExecBaseParams *exec_params, v
 DNNReturnType ff_init_async_attributes(DNNAsyncExecModule *async_module);
 DNNReturnType ff_destroy_async_attributes(DNNAsyncExecModule *async_module);
 DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_module);
+DNNAsyncStatusType dnn_get_async_result(Queue *task_queue, AVFrame **in, AVFrame **out);
```
guoyejun: function name ff_*,

ff_ is needed to avoid conflict with other libraries when libavfilter.a is used.

libavfilter/dnn/dnn_backend_common.c
```diff
@@ -69,3 +69,66 @@ DNNReturnType ff_dnn_fill_task(TaskItem *task, DNNExecBaseParams *exec_params, v
...
+
+static void *async_thread_routine(void *args)
+{
+    DNNAsyncExecModule *async_module = args;
+    void *request = async_module->args;
+
+    async_module->start_inference(request);
+    async_module->callback(request);
+#if HAVE_PTHREAD_CANCEL
+    pthread_exit(0);
```
guoyejun: i'm not sure if pthread_exit is needed here for detached thread, just google it and do not find the result.

guoyejun:
lavfi/dnn: Documentation for Async Module

this commit can be merged with the first commit, they should be together.

run 'git rebase -i HEAD~10' and you can reorder/fixup the commit

libavfilter/dnn/dnn_backend_tf.c
guoyejun: > Yes, the name `InferenceItem` is a bit misleading. The name `SubTaskItem`, as you suggested, seems good for naming since each `InferenceItem` is a part of a `TaskItem`.

@Semmer2 any comment? if no, we can change to SubtaskItem, after shubhanshu02's patches, which reply on InferenceItem, are merged.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -126,6 +139,38 @@ static TFInferRequest *tf_create_inference_request(void)
...
+        return DNN_ERROR;
+    }
+
+    TF_SessionRun(tf_model->session, NULL,
+                  infer_request->tf_input, &infer_request->input_tensor, 1,
+                  infer_request->tf_outputs, infer_request->output_tensors,
+                  task->nb_output, NULL, 0, NULL,
+                  tf_model->status);
+    if (TF_GetCode(tf_model->status) != TF_OK) {
+        av_log(&tf_model->ctx, AV_LOG_ERROR, "Failed to run session when executing model\n");
```
guoyejun: the error message can queried with function TF_Message.

you may add the change here in this commit, or add a new commit for all the cases.

libavfilter/dnn/dnn_backend_tf.c
guoyejun: or LastLevelTaskItem?

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -172,6 +172,24 @@ static DNNReturnType tf_start_inference(void *args)
     return DNN_SUCCESS;
 }
 
+/**
+ * Free the TFRequestItem completely.
+ *
+ * @param arg Address of the TFInferRequest instance.
+ */
+static inline void delete_request_item(TFRequestItem **arg) {
```
guoyejun: for the function name, i don't see delete in ffmpeg, we may use free/destory

libavfilter/dnn/dnn_backend_common.h
```diff
@@ -148,4 +148,20 @@ DNNReturnType ff_dnn_start_inference_async(void *ctx, DNNAsyncExecModule *async_
...
+ * @param exec_params pointer to execution parameters
+ * @param backend_model void pointer to the backend model
+ * @param input_height height of input frame
+ * @param input_width width of input frame
+ * @param ctx pointer to the backend context
+ *
+ * @retval DNN_SUCCESS if successful
+ * @retval DNN_ERROR if allocation fails
+ */
+DNNReturnType dnn_get_output(TaskItem *task, DNNExecBaseParams *exec_params, void *backend_model, int input_height, int input_width, void *ctx);
```
guoyejun: function name: ff_*

and, dnn_get_output is not correct, since we just fill the task here and we do not get output yet, need a new function name.

libavfilter/dnn/dnn_backend_common.c
```diff
@@ -69,3 +69,66 @@ DNNReturnType ff_dnn_fill_task(TaskItem *task, DNNExecBaseParams *exec_params, v
...
+}
+
+DNNReturnType ff_init_async_attributes(DNNAsyncExecModule *async_module)
+{
+    if (!async_module) {
+        return DNN_ERROR;
+    }
+#if HAVE_PTHREAD_CANCEL
+    pthread_attr_init(&async_module->thread_attr);
+    pthread_attr_setdetachstate(&async_module->thread_attr, PTHREAD_CREATE_DETACHED);
```
guoyejun: not sure if the two functions are supported on Windows.

guoyejun: I think no, see https://github.com/FFmpeg/FFmpeg/blob/master/compat/w32pthreads.h for the supported functions for windows.

if no, we might consider the default thread.

guoyejun: @xuguangxin do we now have windows build in CI?  I heard that we have enabled the windows build, if so, could you please have someone to try https://github.com/intel-media-ci/ffmpeg/pull/430, thanks.

libavfilter/dnn/queue.h
shubhanshu02: Actually on adding `queue.h` in `dnn_backend_common.h`, the `make` command gave error as follows:
```
In file included from libavfilter/dnn/dnn_backend_common.h:27,
                 from libavfilter/dnn/dnn_backend_common.c:24:
libavfilter/dnn/queue.h:47:1: error: unknown type name ‘size_t’
   47 | size_t ff_queue_size(Queue *q);
      | ^~~~~~
libavfilter/dnn/queue.h:1:1: note: ‘size_t’ is defined in header ‘<stddef.h>’; did you forget to ‘#include <stddef.h>’?
  +++ |+#include <stddef.h>
    1 | /*
```

libavfilter/dnn/dnn_backend_common.h
shubhanshu02: Oh, that's the reason why all functions have "ff" in the beginning :open_mouth:. Sure, I'll correct it.

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Okay, I'll correct this.

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Sure, I'll add the change here. Should I query the message using TF_Message for other lines as well where status != OK?

libavfilter/dnn/dnn_backend_common.h
shubhanshu02: Yes, it allocates memory for the frames and fills the task. How about `ff_dnn_allocate_frame_and_fill_tasks`?

libavfilter/dnn/dnn_backend_common.c
xuguangxin: No, not yet. Windows CI is harder than we expected. If you just want check it's break windows or not. you can ask fan1x.li@intel.com for help.

libavfilter/dnn/dnn_backend_common.h
guoyejun: let's make it more straightforward, how about ff_dnn_file_gettingoutput_task

libavfilter/dnn/dnn_backend_tf.c
guoyejun: yes, just pending these changes for other cases, let's focus this patch set first

libavfilter/dnn/queue.h
guoyejun: ok

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Sure

libavfilter/dnn/dnn_backend_common.h
shubhanshu02: Yes, looks good. Thanks

libavfilter/dnn/dnn_backend_common.c
shubhanshu02: Not necessary. According to [this](https://man7.org/linux/man-pages/man3/pthread_create.3.html), this is equivalent to `return 0`.


> The new thread terminates in one of the following ways:
> 
> 1. It calls pthread_exit(3), specifying an exit status value that is > available to another thread in the same process that calls > pthread_join(3).
> 2. It returns from start_routine().  This is equivalent to calling pthread_exit(3) with the value supplied in the return statement.

I didn't know FFmpeg works on windows as well, so I included it. I'll remove it. Thanks

libavfilter/dnn/dnn_backend_common.c
guoyejun: checked with Fan, HAVE_PTHREAD_CANCEL is not enabled in Windows, what enabled is PTHREAD_CANCEL_ENABLE.

@shubhanshu02  to make the async support available for future extending in Windows (we can define HAVE_PTHREAD_CANCEL for windows whenever we want to support async), we'd better do not use the functionalities not defined in https://github.com/FFmpeg/FFmpeg/blob/master/compat/w32pthreads.h 

in other words, we could not use PTHREAD_CREATE_DETACHED, we may do phtread_join at the beginning of ff_dnn_execute_model_async_tf or get_async functions (actually, you may find a better place in your code) .

shubhanshu02: Sure, I'll correct this part.

I think joining the thread in the function `ff_dnn_start_inference_async` so as to keep the threading logic separate in the `dnn_backend_common`. 

One question, do we have to check PTHREAD_CANCEL_ENABLE in addition to HAVE_PTHREAD_CANCEL in the code where we use pthread functions?

guoyejun: sure, it is a better place, just to make sure the last thread is joined.

we now just check HAVE_PTHREAD_CANCEL , to make the thing simple.

shubhanshu02: Okay, thanks. I'll update the patches and let you know.

shubhanshu02: Corrected.

libavfilter/dnn/dnn_backend_common.c
```diff
@@ -69,3 +69,57 @@ DNNReturnType ff_dnn_fill_task(TaskItem *task, DNNExecBaseParams *exec_params, v
...
+{
+    int ret;
+
+    if (!async_module) {
+        av_log(ctx, AV_LOG_ERROR, "async_module is null when starting async inference.\n");
+        return DNN_ERROR;
+    }
+
+#if HAVE_PTHREAD_CANCEL
+    pthread_join(async_module->thread_id, NULL);
```
guoyejun: for the firs time that async_module->thread_id has not been created via pthread_create, will this function cause error?

libavfilter/dnn/dnn_backend_common.c
```diff
@@ -69,3 +69,57 @@ DNNReturnType ff_dnn_fill_task(TaskItem *task, DNNExecBaseParams *exec_params, v
...
+    ret = pthread_create(&async_module->thread_id, NULL, async_thread_routine, async_module);
+    if (ret != 0) {
+        av_log(ctx, AV_LOG_ERROR, "Unable to start async inference.\n");
+        return DNN_ERROR;
+    }
+#else
+    av_log(ctx, AV_LOG_WARNING, "pthreads not supported. Roll back to sync.\n");
```
guoyejun: we will see this log tons of times in this path, we'd output this message once (or just several times)  in the whole process. We may output this warning message at another place.

libavfilter/dnn/dnn_backend_common.c
shubhanshu02: No, this won't be a problem. pthread_join returns 3 (no such process) in case no thread with this id is running. I also tried running inference with this patchset and found to be working fine

libavfilter/dnn/dnn_backend_common.c
shubhanshu02: Sure. How about showing this warning in the ff_dnn_load_model_tf function?

guoyejun: looks good

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -116,6 +123,12 @@ static void tf_free_request(TFInferRequest *request)
...
 
+/**
+ * Create a TensorFlow inference request. All properties
+ * are initially unallocated and set as NULL.
+ *
+ * @return pointer to the allocated TFInferRequest instance.
+ */
 static TFInferRequest *tf_create_inference_request(void)
 {
     TFInferRequest *infer_request = av_malloc(sizeof(TFInferRequest));
```
guoyejun, line 138: we forget to check the result, just fix it in another patch.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -898,6 +902,15 @@ DNNModel *ff_dnn_load_model_tf(const char *model_filename, DNNFunctionType func_
         goto err;
     }
 
+    tf_model->task_queue = ff_queue_create();
+    if (!tf_model->task_queue) {
+        goto err;
+    }
+
+#if !HAVE_PTHREAD_CANCEL
+    av_log(ctx, AV_LOG_WARNING, "pthreads not supported. Roll back to sync.\n");
```
guoyejun: #if !HAVE_PTHREAD_CANCEL
if (ctx->options.async)
   av_log(...)
#endif

guoyejun:
I don't have other comments, please fix the two issues and send the 8 (one new patch) patches to community. @Semmer2 please help to verify the patches once they are in the mail list, by running sr/derain/dnn_processing/dnn_detect with tf backend.

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: The async is in the DnnContext which isn't being passed to the load_model function. I also found this line in `dnn_filter_common.c`:
```c
int ff_dnn_init(DnnContext *ctx, DNNFunctionType func_type, AVFilterContext *filter_ctx)
{
...
#if !HAVE_PTHREAD_CANCEL
    if (ctx->async) {
        ctx->async = 0;
        av_log(filter_ctx, AV_LOG_WARNING, "pthread is not supported, roll back to sync.\n");
    }
#endif
...
}
```
So there is no need for checking this for now :thinking:, but I think we'll need this line in the unification patches. Shall I remove this line for now?

libavfilter/dnn/dnn_backend_tf.c
shubhanshu02: Sure, thanks.

shubhanshu02: Sure, thanks.

libavfilter/dnn/dnn_backend_tf.c
guoyejun: yes, as for now, we may remove it, thanks.

shubhanshu02: Thanks, sent.

shubhanshu02:
@guoyejun sir, I have added the async status handling patch (b2e78c1a4bb98b00975777baab112ab12a46e4f9) to this pull request. Please have a look at it. 

Basically, it checks the status of the previously running thread for the same request and if it is `(void *)-1` (named `DNN_ASYNC_FAIL` in this case), the execution was not successful. The status would be 3 in case no such thread was running.
In case of failure during execution, the `async_module->start_inference` returns `DNN_ERROR`, cleans up the execution parameters, and pushes the request back to the queue. On receipt of `DNN_ERROR`, the thread returns with the above exit value without calling the completion callback. So, the possible memory leak was handled in the `start_inference` function.

libavfilter/dnn/dnn_backend_common.c
```diff
@@ -101,14 +106,19 @@ DNNReturnType ff_dnn_async_module_cleanup(DNNAsyncExecModule *async_module)
...
 
     if (!async_module) {
         av_log(ctx, AV_LOG_ERROR, "async_module is null when starting async inference.\n");
         return DNN_ERROR;
     }
 
 #if HAVE_PTHREAD_CANCEL
-    pthread_join(async_module->thread_id, NULL);
+    pthread_join(async_module->thread_id, &status);
+    if (status == DNN_ASYNC_FAIL) {
```
guoyejun, line 123: man pthread_join:
On success, pthread_join() returns 0; on error, it returns an error number.

can we just compare it with -1?

libavfilter/dnn/dnn_backend_common.c
```diff
@@ -79,9 +82,11 @@ static void *async_thread_routine(void *args)
     DNNAsyncExecModule *async_module = args;
     void *request = async_module->args;
 
-    async_module->start_inference(request);
+    if (async_module->start_inference(request) != DNN_SUCCESS) {
+        return DNN_ASYNC_FAIL;
+    }
     async_module->callback(request);
-    return NULL;
+    return DNN_ASYNC_SUCCESS;
```
guoyejun, line 89: the return type is "void *", and where do we check the return value of this function?

libavfilter/dnn/dnn_backend_common.c
shubhanshu02: The error number is when it cannot join the thread. In our case, the thread will join successfully and we are catching the status of `async_thread_routine()`. If we change the value of `DNN_ASYNC_SUCCESS` to something like `(void *)40`, then on the success the status will be 40 but the return value of `pthread_join()` will be 0. 

For example, the output of the below code is `0 -1`.

```C
void *runner(void *arg)
{
    sleep(1);
    return (void *)-1;
}

int main() {
    pthread_t thread;
    int returnValue = pthread_create(&thread, NULL, runner, NULL);
    void *status = 0;
    int ret = pthread_join(thread, &status);
    printf("%d %d\n", ret, (int *)status);
}
```

guoyejun: thanks, i see, looks good.

libavfilter/dnn/dnn_backend_common.c
shubhanshu02: Yes, on line 26 I have defined this status value as follows. The return value is checked on Line 117 before starting the next inference.

```C
#define DNN_ASYNC_SUCCESS (void *)0
```

libavfilter/dnn/dnn_backend_common.c
shubhanshu02: Just one question, do we need to check the exit status in `ff_dnn_async_module_cleanup` also? It is called only when the request item is destroyed to ensure the thread was joined.

guoyejun: yes, i think so, we may output a warning here

shubhanshu02: Okay, I'll add this to the patch. Thank you.

guoyejun:
looks good to me, and any other change beside the last patch?

 you may send the v3 patches together with is patch,  since they are nature to be together.

@Semmer2 please help to verify the patches.

shubhanshu02:
> other change beside the last patch?

I can't think of anything else to make the patchset better for now. I think that's all for this patchset. 
I'll send the v3 patchset in a few minutes. (Update: I have sent them)
Thank you.

shubhanshu02:
Thank you for reviewing the pull request, @guoyejun sir.

PR #409, opened 2021-05-31
3dinput

wenbinc-Bin:
Add input frame from and to vulkan for ffmpeg-vaapi and ffmpeg-qsv.

### DIFF
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -1238,6 +1238,10 @@ static void clear_unused_frames(QSVEncContext *q)
     while (cur) {
         if (cur->used && !cur->surface.Data.Locked) {
             free_encoder_ctrl_payloads(&cur->enc_ctrl);
+            if (cur->external_frame) {
+                av_freep(&cur->surface.Data.MemId);
+                cur->external_frame = 0;
+            }
             if (cur->frame->format == AV_PIX_FMT_QSV) {
                 av_frame_unref(cur->frame);
             }
@@ -1304,12 +1308,20 @@ static int submit_frame(QSVEncContext *q, const AVFrame *frame,
 
         qf->surface = *(mfxFrameSurface1*)qf->frame->data[3];
 
+
         if (q->frames_ctx.mids) {
             ret = ff_qsv_find_surface_idx(&q->frames_ctx, qf);
-            if (ret < 0)
-                return ret;
-
-            qf->surface.Data.MemId = &q->frames_ctx.mids[ret];
+            if (ret >= 0)
+                qf->surface.Data.MemId = &q->frames_ctx.mids[ret];
+        }
+        if (!q->frames_ctx.mids || ret < 0) {
+            QSVMid *mid = NULL;
+            mid = (QSVMid *)av_mallocz(sizeof(*mid));
+            if (!mid)
+                return AVERROR(ENOMEM);
+            mid->handle = qf->surface.Data.MemId;
+            qf->surface.Data.MemId = mid;
+            qf->external_frame = 1;
         }
     } else {
         /* make a copy if the input is not padded as libmfx requires */
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -456,8 +456,8 @@ static int init_video_param_jpeg(AVCodecContext *avctx, QSVEncContext *q)
     if (avctx->hw_frames_ctx) {
         AVHWFramesContext *frames_ctx    = (AVHWFramesContext *)avctx->hw_frames_ctx->data;
         AVQSVFramesContext *frames_hwctx = frames_ctx->hwctx;
-        q->param.mfx.FrameInfo.Width  = frames_hwctx->surfaces[0].Info.Width;
-        q->param.mfx.FrameInfo.Height = frames_hwctx->surfaces[0].Info.Height;
+        q->param.mfx.FrameInfo.Width  = frames_hwctx->reserve_surface.Info.Width;
+        q->param.mfx.FrameInfo.Height = frames_hwctx->reserve_surface.Info.Height;
     }
 
     if (avctx->framerate.den > 0 && avctx->framerate.num > 0) {
@@ -577,8 +577,8 @@ static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
     if (avctx->hw_frames_ctx) {
         AVHWFramesContext *frames_ctx = (AVHWFramesContext*)avctx->hw_frames_ctx->data;
         AVQSVFramesContext *frames_hwctx = frames_ctx->hwctx;
-        q->param.mfx.FrameInfo.Width  = frames_hwctx->surfaces[0].Info.Width;
-        q->param.mfx.FrameInfo.Height = frames_hwctx->surfaces[0].Info.Height;
+        q->param.mfx.FrameInfo.Width  = frames_hwctx->reserve_surface.Info.Width;
+        q->param.mfx.FrameInfo.Height = frames_hwctx->reserve_surface.Info.Height;
     }
 
     if (avctx->framerate.den > 0 && avctx->framerate.num > 0) {
--- a/libavutil/hwcontext_qsv.c
+++ b/libavutil/hwcontext_qsv.c
@@ -1025,40 +1025,106 @@ static int qsv_frames_derive_to(AVHWFramesContext *dst_ctx,
     return 0;
 }
 
+static void qsv_umap_from_vaapi(AVHWFramesContext *dst_fc,
+                                 HWMapDescriptor *hwmap)
+{
+    mfxFrameSurface1 *new_sur = (mfxFrameSurface1 *)hwmap->priv;
+    av_freep(&new_sur->Data.MemId);
+    av_freep(&new_sur);
+}
+
 static int qsv_map_to(AVHWFramesContext *dst_ctx,
                       AVFrame *dst, const AVFrame *src, int flags)
 {
     AVQSVFramesContext *hwctx = dst_ctx->hwctx;
+    mfxFrameSurface1 *new_sur = NULL;
+    VASurfaceID *new_vasur = NULL;
+
     int i, err;
 
     for (i = 0; i < hwctx->nb_surfaces; i++) {
 #if CONFIG_VAAPI
-        if (*(VASurfaceID*)hwctx->surfaces[i].Data.MemId ==
-            (VASurfaceID)(uintptr_t)src->data[3])
-            break;
+        if (src->format == AV_PIX_FMT_VAAPI) {
+            if (*(VASurfaceID*)hwctx->surfaces[i].Data.MemId ==
+                (VASurfaceID)(uintptr_t)src->data[3])
+                break;
+        } else
 #endif
 #if CONFIG_DXVA2
-        if ((IDirect3DSurface9*)hwctx->surfaces[i].Data.MemId ==
-            (IDirect3DSurface9*)(uintptr_t)src->data[3])
-            break;
+        if (src->format == AV_PIX_FMT_DXVA2_VLD) {
+            if ((IDirect3DSurface9*)hwctx->surfaces[i].Data.MemId ==
+                (IDirect3DSurface9*)(uintptr_t)src->data[3])
+                break;
+        } else
 #endif
+        {
+            return AVERROR(ENOSYS);
+        }
     }
     if (i >= hwctx->nb_surfaces) {
-        av_log(dst_ctx, AV_LOG_ERROR, "Trying to map from a surface which "
-               "is not in the mapped frames context.\n");
-        return AVERROR(EINVAL);
-    }
+        switch (src->format) {
+#if CONFIG_VAAPI
+        case AV_PIX_FMT_VAAPI:
+        {
+            new_sur = (mfxFrameSurface1 *)av_mallocz(sizeof(*new_sur));
+            if (!new_sur) {
+                err = AVERROR(ENOMEM);
+                goto qsv_map_to_err;
+            }
+            err = qsv_init_surface(dst_ctx, new_sur);
+            if (err < 0)
+                goto qsv_map_to_err;
 
-    err = ff_hwframe_map_create(dst->hw_frames_ctx,
-                                dst, src, NULL, NULL);
-    if (err)
-        return err;
+            new_vasur = (VASurfaceID *)av_mallocz(sizeof(*new_vasur));
+            if (!new_vasur) {
+                err = AVERROR(ENOMEM);
+                goto qsv_map_to_err;
+            }
+            *new_vasur = (VASurfaceID)(uintptr_t)src->data[3];
+            new_sur->Data.MemId = new_vasur;
+
+            err = ff_hwframe_map_create(dst->hw_frames_ctx, dst, src,
+                                        &qsv_umap_from_vaapi,
+                                        (void*)new_sur);
+            if (err)
+                goto qsv_map_to_err;
+
+            av_log(dst_ctx, AV_LOG_DEBUG, "Trying to map from a surface which "
+                "is not in the mapped frames context, so create a new surface\n");
+        }
+        break;
+#endif
+#if CONFIG_DXVA2
+        case AV_PIX_FMT_DXVA2_VLD:
+        {
+            av_log(dst_ctx, AV_LOG_ERROR, "Trying to map from a surface which "
+                "is not in the mapped frames context.\n");
+            return AVERROR(EINVAL);
+        }
+        break;
+#endif
+        default:
+            return AVERROR(ENOSYS);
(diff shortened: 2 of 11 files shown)


wenbinc-Bin:
@xuguangxin @xhaihao @feiwan1 

Could you please help me to review these patches.

Thanks
Wenbin

wenbinc-Bin:
@xuguangxin @xhaihao @feiwan1 
Is there any feedback on this patchset?
Thank you

libavutil/hwcontext_vulkan.c
```diff
@@ -1349,7 +1349,7 @@ static int vulkan_device_init(AVHWDeviceContext *ctx)
         }
     }
 
-    err = load_functions(ctx, 1, 0);
```
xuguangxin: why the original author, not use 1?
any side effects for this change?

libavutil/hwcontext_vulkan.c
```diff
@@ -1448,6 +1458,8 @@ static int vulkan_device_derive(AVHWDeviceContext *ctx,
         if (strstr(vendor, "AMD"))
             dev_select.vendor_id = 0x1002;
 
+        av_dict_set_int(&opts, "single_linear", 1, NULL);
```
xuguangxin: will this impact AMD? could we just use single_linear for intel only?

libavutil/hwcontext_vulkan.c
```diff
@@ -1659,6 +1659,88 @@ static void vulkan_frame_free(void *opaque, uint8_t *data)
     av_free(f);
 }
 
+static int alloc_bind_single_mem(AVHWFramesContext *hwfc, AVVkFrame *f,
```
xuguangxin: nearly same code as alloc_bind_mem, could we pass one flag to 
static int alloc_bind_mem(AVHWFramesContext *hwfc, AVVkFrame *f,    void *alloc_pnext, size_t alloc_pnext_stride, int use_single_linear_images )

libavutil/hwcontext_vulkan.c
```diff
@@ -2258,16 +2258,31 @@ static int vulkan_map_frame_to_mem(AVHWFramesContext *hwfc, AVFrame *dst,
...
+        ret = vk->MapMemory(hwctx->act_dev, f->mem[0], 0,
+                            VK_WHOLE_SIZE, 0, (void **)&mem_addr);
         if (ret != VK_SUCCESS) {
             av_log(hwfc, AV_LOG_ERROR, "Failed to map image memory: %s\n",
                 vk_ret2str(ret));
             err = AVERROR_EXTERNAL;
             goto fail;
         }
-        mapped_mem_count++;
+        for (int i = 0; i < planes; i++) {
```
xuguangxin, line 2232: missed mapped_mem_count?
How about 
```
loop = use_single_linear_images ? 1 :planes

for (int i = 0; i < loop; i++)
    xxxx..
if (use_single_linear_images) {
   for (int i = i; i < planes; i++) {
         dst->data[i] = dst->data[0]+ f->offset[i];
   }
}
```

libavutil/hwcontext_vulkan.c
```diff
@@ -1659,6 +1659,88 @@ static void vulkan_frame_free(void *opaque, uint8_t *data)
...
+        req.sType = VK_STRUCTURE_TYPE_MEMORY_REQUIREMENTS_2;
+        req.pNext = &ded_req;
+
+        vk->GetImageMemoryRequirements2(hwctx->act_dev, &req_desc, &req);
+
+        if (f->tiling == VK_IMAGE_TILING_LINEAR)
+            req.memoryRequirements.size = FFALIGN(req.memoryRequirements.size,
+                                                  p->props.properties.limits.minMemoryMapAlignment);
+
+        mem_size_list[i] += FFALIGN(req.memoryRequirements.size/hwfc->width, 64) * hwfc->width;
```
xuguangxin: Where is the 64 come from? in the future platforms it may change.
Could we get the 64 from somewhere?

fftools/ffmpeg_qsv.c
```diff
@@ -48,6 +48,22 @@ static int qsv_device_init(InputStream *ist)
...
+
+    if (ist->hwaccel_device) {
+        hw_device = hw_device_get_by_name(ist->hwaccel_device);
+        if (hw_device && hw_device->type == AV_HWDEVICE_TYPE_QSV) {
+            hw_device_ctx = av_buffer_ref(hw_device->device_ref);
+            if (!hw_device_ctx)
+                err = AVERROR(ENOMEM);
+            else
+                err = 0;
+            goto err_out;
```
xuguangxin: why we need err_out for err = 0

libavcodec/qsvenc.c
```diff
@@ -1304,12 +1312,18 @@ static int submit_frame(QSVEncContext *q, const AVFrame *frame,
...
         if (q->frames_ctx.mids) {
             ret = ff_qsv_find_surface_idx(&q->frames_ctx, qf);
-            if (ret < 0)
-                return ret;
-
-            qf->surface.Data.MemId = &q->frames_ctx.mids[ret];
+            if (ret >= 0)
+                qf->surface.Data.MemId = &q->frames_ctx.mids[ret];
+        }
+        if (!q->frames_ctx.mids || ret < 0) {
```
xuguangxin, line 1317: if q->frames_ctx.mids is not NULL, ret < 0 is a error.

libavcodec/qsvenc.c
```diff
@@ -1304,12 +1312,18 @@ static int submit_frame(QSVEncContext *q, const AVFrame *frame,
...
+            QSVMid *mid = NULL;
+            mid = (QSVMid *)av_mallocz(sizeof(*mid));
```
xuguangxin, line 1319: need check mid

libavutil/hwcontext_qsv.c
```diff
@@ -1044,9 +1046,14 @@ static int qsv_map_to(AVHWFramesContext *dst_ctx,
 #endif
     }
     if (i >= hwctx->nb_surfaces) {
-        av_log(dst_ctx, AV_LOG_ERROR, "Trying to map from a surface which "
-               "is not in the mapped frames context.\n");
-        return AVERROR(EINVAL);
+        new_sur = (mfxFrameSurface1 *)av_mallocz(sizeof(*new_sur));
```
xuguangxin: need check the malloc

libavutil/hwcontext_vulkan.c
wenbinc-Bin: I don't know, author didn't mention in his patch. But without using 1, there will be an segmentation error because vk->CreateCommandPool is not load. 
Segmentation error occur in this line:
https://github.com/FFmpeg/FFmpeg/blob/2fac1e370c9cf48b2e9d4a7f2c0d7236017f9bbe/libavutil/hwcontext_vulkan.c#L1018

xuguangxin: Without any of you patch, it will not have a crash, right?

libavcodec/qsvenc.c
wenbinc-Bin: if mids is not NULL, ret is the return value from ff_qsv_find_surface_idx(), so if ret >=0 we skip this branch, and if ret < 0, it means the surface is not in the pool, and we need go into this branch to create new resource for this frame.

xuguangxin: in some error cases, mids are not null, and ret will < 0 too. In this case we can't create a new resource for it, we need to report an error.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: without this patch it will crash. without other pathes, the command line will report error.

libavcodec/qsvenc.c
wenbinc-Bin: Ok, I got your point.
The ff_qsv_find_surface_idx() is a simple function, It returns minus value only when the surface is not found, so I don't category the returned value.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: Good advice, and now change it.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: I've merged them into one function.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: Good advice, I've changed it.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: Here I made a mistake, I want the offset to be multiple of width.

fftools/ffmpeg_qsv.c
wenbinc-Bin: Because "err_out" release the resource, I want to make sure the resource is released. However, I found at this line, the resource is not allocated yet, so it is unnecessary. Now I changed it. Thanks.

libavcodec/qsvenc.c
wenbinc-Bin: Add check, thanks.

libavutil/hwcontext_qsv.c
wenbinc-Bin: Add check, thanks.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: I submit a ticket. https://trac.ffmpeg.org/ticket/9279#ticket.
Temporarily use this patch before they fix the issue.

libavutil/hwcontext_vulkan.c
```diff
@@ -2188,6 +2188,7 @@ static int vulkan_map_frame_to_mem(AVHWFramesContext *hwfc, AVFrame *dst,
 {
     VkResult ret;
     int err, mapped_mem_count = 0;
+    int loop = 0;
```
xuguangxin: init is not needed

fftools/ffmpeg_qsv.c
```diff
@@ -48,6 +48,21 @@ static int qsv_device_init(InputStream *ist)
 {
     int err;
     AVDictionary *dict = NULL;
+    HWDevice *hw_device = NULL;
+
+    if (ist->hwaccel_device) {
+        hw_device = hw_device_get_by_name(ist->hwaccel_device);
+        if (hw_device && hw_device->type == AV_HWDEVICE_TYPE_QSV) {
```
xuguangxin: maye you can use
if (hw_device) {
 if (hw_device->type != ..)


}

fftools/ffmpeg_qsv.c
```diff
@@ -48,6 +48,21 @@ static int qsv_device_init(InputStream *ist)
...
+            hw_device_ctx = av_buffer_ref(hw_device->device_ref);
+            if (!hw_device_ctx)
+                return AVERROR(ENOMEM);
+            else
```
xuguangxin: else is not needed.

fftools/ffmpeg_qsv.c
```diff
@@ -48,6 +48,21 @@ static int qsv_device_init(InputStream *ist)
 {
     int err;
     AVDictionary *dict = NULL;
+    HWDevice *hw_device = NULL;
+
+    if (ist->hwaccel_device) {
+        hw_device = hw_device_get_by_name(ist->hwaccel_device);
+        if (hw_device && hw_device->type == AV_HWDEVICE_TYPE_QSV) {
+            hw_device_ctx = av_buffer_ref(hw_device->device_ref);
```
xuguangxin: who will release this?

libavutil/hwcontext_vulkan.c
```diff
@@ -1696,6 +1699,23 @@ static int alloc_bind_mem(AVHWFramesContext *hwfc, AVVkFrame *f,
...
+                return AVERROR(EINVAL);
+            }
+            if (memory_requirements.size == 0) {
+                memory_requirements = req.memoryRequirements;
+            } else if (memory_requirements.memoryTypeBits != req.memoryRequirements.memoryTypeBits) {
+                av_log(hwfc, AV_LOG_ERROR, "the param for each planes are not the same\n");
+                return AVERROR(EINVAL);
+            }
+
+            mem_size_list[i] += (req.memoryRequirements.size/hwfc->width + 1) * hwfc->width;
```
xuguangxin, line 1714: why not use req.memoryRequirements.size?
is this workable for other format like i420?

libavutil/hwcontext_qsv.c
```diff
@@ -1044,9 +1046,20 @@ static int qsv_map_to(AVHWFramesContext *dst_ctx,
...
+        if (!new_sur)
+            return AVERROR(ENOMEM);
+        new_vasur = (VASurfaceID *)av_mallocz(sizeof(VASurfaceID));
+        if (!new_vasur)
+            return AVERROR(ENOMEM);
```
xuguangxin: this return will leak new_sur

libavutil/hwcontext_vulkan.c
wenbinc-Bin: change it, thanks.

fftools/ffmpeg_qsv.c
wenbinc-Bin: change it, thanks

fftools/ffmpeg_qsv.c
wenbinc-Bin: change it, thanks

fftools/ffmpeg_qsv.c
wenbinc-Bin: Add unref operation in qsv_uninit(). It miss an unref even without this patchset. Seems it is a bug for ffmpeg-qsv.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: Because req.memoryRequirements.size is not perfectly aligned with width, and libva require that.
It should work for other format as well and I will test it.

libavutil/hwcontext_qsv.c
wenbinc-Bin: change it, thanks.

xhaihao (commented):
Someone disagree with the change of hwaccel_device, see http://ffmpeg.org/pipermail/ffmpeg-devel/2020-September/269074.html for detail

wenbinc-Bin:
> Someone disagree with the change of hwaccel_device, see http://ffmpeg.org/pipermail/ffmpeg-devel/2020-September/269074.html for detail

I read the thread line...  According to softworkz's first reply, he seems to agree to use -hwaccel_device to refer to an existing device created with -init_hw_device by name. Maybe he can accept this patch?

I see mypopy disagree to mix qsv and vaapi in one ffmpeg pipeline, which is what I am going to do...

xhaihao:
> > Someone disagree with the change of hwaccel_device, see http://ffmpeg.org/pipermail/ffmpeg-devel/2020-September/269074.html for detail
> 
> I read the thread line... According to softworkz's first reply, he seems to agree to use -hwaccel_device to refer to an existing device created with -init_hw_device by name. Maybe he can accept this patch?
> 

According to the feedback in this and other threads, we should support HW_DEVICE_CTX and remove AD_HOC for qsv decoder. If so, the device created by init_hw_device can be used for qsv decoder, like as qsv encoder. I am looking into it. Does it block your implementation for "vaapi decoding + qsv encoding" ? 

> I see mypopy disagree to mix qsv and vaapi in one ffmpeg pipeline, which is what I am going to do...

Yes, he disagreed with the combination of vaapi and qsv. You may submit your patches to see whether there are more comments.

wenbinc-Bin:
> According to the feedback in this and other threads, we should support HW_DEVICE_CTX and remove AD_HOC for qsv decoder. If so, the device created by init_hw_device can be used for qsv decoder, like as qsv encoder. I am looking into it. Does it block your implementation for "vaapi decoding + qsv encoding" ?

I think it doesn't block me.
> 
> > I see mypopy disagree to mix qsv and vaapi in one ffmpeg pipeline, which is what I am going to do...
> 
> Yes, he disagreed with the combination of vaapi and qsv. You may submit your patches to see whether there are more comments.

Ok, thanks Haihao.

wenbinc-Bin:
I see vulkan init error is already fixed, so I remove one of my patches.

wenbinc-Bin:
After Haihao mentioned the discussion about -hwaccel_device, I check the code and found that the using "-hwaccel_device qs" to choose the device_ctx for -hwaccel qsv may not comply the logic of ffmpeg_hw, so I remove that patch and add a new patch to support deriving vaapi device_ctx from qsv to implemnet 3Dinput function.

wenbinc-Bin:
On DG2 libva is not compatible with vulkan. Will reopen this pr when this problem is fixed on driver side.

PR #420, opened 2021-06-30
AVX512 Optimization for Gaussian Filter 2D

iqsxw:
ff_postscale_slice_c: 99
ff_postscale_slice_avx2: 5000
ff_postscale_slice_avx512: 5000

ff_horiz_slice_avx2(old): 109
ff_horiz_slice_avx2(new): 666.66
ff_horiz_slice_avx512: 1000

ff_verti_slice_c: 32.6
ff_verti_slice_avx2: 476.2
ff_verti_slice_avx512: 833.3


libavfilter/x86/vf_gblur.asm
```diff
@@ -20,17 +20,371 @@
...
 
 %macro HORIZ_SLICE 0
-%if UNIX64
+%if cpuflag(avx512)
+cglobal horiz_slice_asm, 5, 12, 32, buffer, width, height, steps, localbuf, x, y, step, stride, remain, ptr
+%elif UNIX64
 cglobal horiz_slice, 4, 9, 9, ptr, width, height, steps, x, y, step, stride, remain
 %else
-cglobal horiz_slice, 4, 9, 9, ptr, width, height, steps, nu, bscale, x, y, step, stride, remain
+cglobal horiz_slice, 4, 9, 9 ptr, width, height, steps, nu, bscale, x, y, step, stride, remain
```
xhaihao: Is this change needed ? it seems you should not remove ',' here.

iqsxw: No, it's not necessary at all. It was supposed to be deleted by me accidentally.

xuguangxin:
@iqsxw great! 
could you invite yanfei and other colleagues to review?
thanks

libavfilter/x86/vf_gblur_init.c
```diff
@@ -51,5 +55,17 @@ av_cold void ff_gblur_init_x86(GBlurContext *s)
...
+        s->horiz_slice = ff_horiz_slice_avx512;
+    }
 #endif
 }
+
+void ff_horiz_slice_avx512(float *buffer, int width, int height, int steps, float nu, float bscale)
+{
+    float *localbuf;
+
+    localbuf = (float *)av_malloc(16 * sizeof(float) * width);
```
xuguangxin: we do not need malloc/free this again and again.
how about we move this malloc to higher level and pass it to this function. 
you can break this to two commits:
1. add temporary buffer to ff_postscale_slice_xxxx
2. implement ff_postscale_slice_avx512

libavfilter/x86/vf_gblur.asm
```diff
@@ -20,17 +20,371 @@
...
+
+            sub ptrq, 4
+            mulps zm3, zm1
+            kxnorw k7, k7, k7
+            vscatterdps [ptrq + zm2*4]{k7}, zm3
+
+            ; Filter leftwards
+            dec xq
+            .loop_x_back:
+                ; Read 16x16 matrix from local buffer
```
xuguangxin: not sure we can use some trick like this  or not
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=transpose&expand=3186,5961,5961

libavfilter/x86/vf_gblur_init.c
iqsxw: It seemed that we have to malloc again and again. The working of  horiz_slice was dispatched by worker queue of multi-threads. How and when each of horiz_slice that was called or done, was unpredictable, so each of horiz_slice needs a unique buffer to save their temp values to keep thread-safe.

libavfilter/x86/vf_gblur.asm
iqsxw: It's already tested that using unpack and shuffle instructions like `_MM_TRANSPOSE4_PS ` is able to transpose 16x16 matrix, but it could decrease executing time 36.8% to use permute instruction to transpose matrix16x16. I think it's worthy to unroll for better performance.

libavfilter/x86/vf_gblur_init.c
xuguangxin: maybe you can use jobnr

HuiLi-Intel:
Great work!

libavfilter/x86/vf_gblur_init.c
iqsxw: Yeah. We might be able to change ThreadData to carry a local buffer variable and slice it use the values of jobnr.
But there are some questions. 
1. If we change ThreadData, which would affect other functions, there would be more things we need to update. When moving the local buffer to a higher level, the interface of functions would change. We might have to keep the function interface identical or we might be not able to pass the fate test.
2. Would it be a problem that I modify so much for one patch?

xuguangxin:
@iqsxw please do not use merge, use "git pull -r "
other things are ok to me

yanfeich:
OK to me.

uartie:
close. checks pass.  please reopen if you need more reviews and pre-check

PR #426, opened 2021-07-15
Qsv video signal info

xhaihao:
1. get the color properties from the SDK for decoder. 
2. pass the color properties to the SDK  for encoder

### DIFF
--- a/libavcodec/qsvdec.c
+++ b/libavcodec/qsvdec.c
@@ -311,7 +311,8 @@ static int qsv_decode_header(AVCodecContext *avctx, QSVContext *q,
                              mfxVideoParam *param)
 {
     int ret;
-
+    mfxExtVideoSignalInfo video_signal_info = { 0 };
+    mfxExtBuffer *header_ext_params[1] = { (mfxExtBuffer *)&video_signal_info };
     mfxBitstream bs = { 0 };
 
     if (avpkt->size) {
@@ -336,6 +337,12 @@ static int qsv_decode_header(AVCodecContext *avctx, QSVContext *q,
         return ret;
 
     param->mfx.CodecId = ret;
+    video_signal_info.Header.BufferId = MFX_EXTBUFF_VIDEO_SIGNAL_INFO;
+    video_signal_info.Header.BufferSz = sizeof(video_signal_info);
+    // The SDK doesn't support other ext buffers, so do not append this buffer
+    // to the existent buffer array
+    param->ExtParam    = header_ext_params;
+    param->NumExtParam = 1;
     ret = MFXVideoDECODE_DecodeHeader(q->session, &bs, param);
     if (MFX_ERR_MORE_DATA == ret) {
        return AVERROR(EAGAIN);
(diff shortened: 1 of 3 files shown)


xhaihao:
@feiwan1 @uartie @xuguangxin @wenbinc-Bin Any comment? I will send it to the mailing list if no objection.

libavcodec/qsvdec.c
```diff
@@ -336,6 +337,10 @@ static int qsv_decode_header(AVCodecContext *avctx, QSVContext *q,
         return ret;
 
     param->mfx.CodecId = ret;
+    video_signal_info.Header.BufferId = MFX_EXTBUFF_VIDEO_SIGNAL_INFO;
+    video_signal_info.Header.BufferSz = sizeof(video_signal_info);
+    param->ExtParam    = header_ext_params;
```
uartie, line 344: Can `param->ExtParam` already contain something here (maybe from `qsv_decode_preinit`)?

xhaihao: Yes, it can contain something here, hence it will be reset to q->ext_buffers after MFXVideoDECODE_DecodeHeader. 
Note MFX_EXTBUFF_VIDEO_SIGNAL_INFO is only used in MFXVideoDECODE_DecodeHeader() and other ext buffer might result in error in MFXVideoDECODE_DecodeHeader().  So we can't append MFX_EXTBUFF_VIDEO_SIGNAL_INFO to param->ExtParam.

libavcodec/qsvdec.c
```diff
@@ -344,6 +349,23 @@ static int qsv_decode_header(AVCodecContext *avctx, QSVContext *q,
         return ff_qsv_print_error(avctx, ret,
                 "Error decoding stream header");
 
+    if (avctx->color_range == AVCOL_RANGE_UNSPECIFIED)
+        avctx->color_range = video_signal_info.VideoFullRange ? AVCOL_RANGE_JPEG : AVCOL_RANGE_MPEG;
+
+    if (video_signal_info.ColourDescriptionPresent) {
+        if (avctx->color_primaries == AVCOL_PRI_UNSPECIFIED)
```
xuguangxin: Will avctx->color_primaries mismatch with video_signal_info.ColourPrimaries, if this is false?

xhaihao: User should have filled the color properties after creating the codec context if  this is false. It's user's responsibility to fill these parameters with right values if he/she want to fill these parameters

xhaihao: According to https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/avcodec.h#L942-L968, these property should be set by libavcodec for decoding, I changed the code to set these properties when video_signal_info.ColourDescriptionPresent is 1.

xhaihao:
Are there more comments about this patch ?

xhaihao:
sent them to the mailing list. 
http://ffmpeg.org/pipermail/ffmpeg-devel/2021-July/282703.html
http://ffmpeg.org/pipermail/ffmpeg-devel/2021-July/282704.html

PR #431, opened 2021-07-29
lavfi/qsvvpp: do not mix up FFmpeg and SDK error code

xhaihao:
The function ff_qsvvpp_filter_frame should return a FFmpeg error code if
there is an error. However it might return a SDK error code without this
patch.


uartie:
LGTM

xhaihao:
See http://ffmpeg.org/pipermail/ffmpeg-devel/2021-July/282797.html

PR #434, opened 2021-08-09
Fix vulkan

wenbinc-Bin:
Try to fix vulkan problem on latest intel driver.


wenbinc-Bin:
Hi @xhaihao @xuguangxin. Any comments on this PR?

libavutil/hwcontext_vulkan.c
```diff
@@ -2342,7 +2342,8 @@ static int vulkan_map_from_drm_frame_desc(AVHWFramesContext *hwfc, AVVkFrame **f
                      hwfc->sw_format, src->width, src->height, i);
 
         for (int j = 0; j < planes; j++) {
-            plane_data[j].offset     = desc->layers[i].planes[j].offset;
+            plane_data[j].offset     =
+                desc->layers[i].planes[j].offset - desc->layers[i].planes[0].offset;
```
xhaihao: AVDRMPlaneDescriptor has the member of object_index, which means different planes may have different objects, so 'desc->layers[i].planes[j].offset - desc->layers[i].planes[0].offset' doesn't make sense to me.

wenbinc-Bin: I am not aware of that. Thank you.
According to the function BindImageMemory2() below, vulkan assume one image (one layer) should be in one drm object. I will add a check here.

libavfilter/vulkan.c
```diff
@@ -472,33 +472,36 @@ int ff_vk_add_exec_dep(AVFilterContext *avctx, FFVkExecContext *e,
...
-        }
+            e->sem_wait[e->sem_wait_cnt] = f->sem[i];
+            e->sem_wait_dst[e->sem_wait_cnt] = in_wait_dst_flag;
+            e->sem_wait_cnt++;
+        } else {
 
-        e->sem_wait[e->sem_wait_cnt] = f->sem[i];
-        e->sem_wait_dst[e->sem_wait_cnt] = in_wait_dst_flag;
-        e->sem_wait_cnt++;
+            e->sem_sig = av_fast_realloc(e->sem_sig, &e->sem_sig_alloc,
```
xuguangxin: Maybe you can send other patch to the realloc issue. 
the ralloc usage is wrong?
when realloc returns NULL, it will not free original memory.  This code will leak memory at that time

libavutil/hwcontext_vulkan.c
```diff
@@ -2016,11 +2033,63 @@ static int vulkan_frames_init(AVHWFramesContext *hwfc)
...
+            .pNext = NULL,
+            .drmFormatModifierCount = 0,
+            .pDrmFormatModifierProperties = NULL,
+        };
+        VkFormatProperties2 prop = {
+            .sType = VK_STRUCTURE_TYPE_FORMAT_PROPERTIES_2,
+            .pNext = &mod_props_list,
+        };
+
+        vk->GetPhysicalDeviceFormatProperties2(dev_hwctx->phys_dev, fmt[0], &prop);
```
xuguangxin, line 2112: better check return value

libavutil/hwcontext_vulkan.c
```diff
@@ -2016,11 +2033,63 @@ static int vulkan_frames_init(AVHWFramesContext *hwfc)
...
+        if (!mod_props_list.drmFormatModifierCount) {
+            av_log(hwfc, AV_LOG_ERROR, "There are not supported modifiers for sw_format\n");
+            return AVERROR(EINVAL);
+        }
+
+        mod_props = (VkDrmFormatModifierPropertiesEXT *)av_mallocz(
```
xuguangxin: check malloc

libavutil/hwcontext_vulkan.c
```diff
@@ -2016,11 +2033,63 @@ static int vulkan_frames_init(AVHWFramesContext *hwfc)
...
+            sizeof(*mod_props) * mod_props_list.drmFormatModifierCount);
+        mod_props_list.pDrmFormatModifierProperties = mod_props;
+        vk->GetPhysicalDeviceFormatProperties2(dev_hwctx->phys_dev, fmt[0], &prop);
```
xuguangxin, line 2121: return

libavutil/hwcontext_vulkan.c
```diff
@@ -2016,11 +2033,63 @@ static int vulkan_frames_init(AVHWFramesContext *hwfc)
...
+
+        hwctx->modifiers = (uint64_t *)av_mallocz(sizeof(*hwctx->modifiers)*
```
xuguangxin: Could we just define a MAX_VULKAN_MODIFIERS to 16.
If it is larger than 16 we return an error at the very beginning. 
using this approach, mod_props[MAX_VULKAN_MODIFIERS ] is a local variable and we do not malloc
AVVulkanFramesContext->modifiers

libavfilter/vulkan.c
wenbinc-Bin: I think the semaphore usage is wrong. You can see the previous code. The semaphores are set to both wait and signal. I don't change its' way to use realloc, I rearrange them to set watSem and signalSem separately. 

The original resource will be free in filter uninit() function. I think program will go into uninit() function if function returning error.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: This function returns void.
Normally we call this function twice. The first time we get number to alloc memory and the second times to get real properties. If we can run successfully at first time we assume the second times must success.
I take this part of code as reference: [https://github.com/intel-media-ci/ffmpeg/blob/835eb0a556d678dd65bb1978519c4d62d93c37c4/libavutil/hwcontext_vulkan.c#L912](https://github.com/intel-media-ci/ffmpeg/blob/835eb0a556d678dd65bb1978519c4d62d93c37c4/libavutil/hwcontext_vulkan.c#L912)

libavutil/hwcontext_vulkan.c
wenbinc-Bin: Same reason as above.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: Always miss it.
Thank you.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: done

libavutil/hwcontext_vulkan.c
wenbinc-Bin: change to use MAX_VULAKN_MODIFIERS. done

libavutil/hwcontext_vulkan.c
wenbinc-Bin: The offset in VkSubresourceLayout and the offset in VkBindImageMemoryInfo is confusing to me. 
According to spec, the plane_data[j].offset is relative to the base address of plane and the drm's offset is used to bind memory below. They are different. I think the offset here should be 0. At least setting to 0 can get correct output. 
I need more investigation.
Thanks for pointing this out .

wenbinc-Bin:
I decide to put some old patches here and pick qsv related patches out as another patchset.

libavfilter/vulkan.c
xuguangxin: uninit can't free the memory since you already set e->sem_sig to NULL.

wenbinc-Bin: You are right. My bad.
Now I have fixed it.
Thanks a lot.

libavutil/hwcontext_vulkan.c
wenbinc-Bin: According to spce, if it is single planer, this is offset to image and it should be zero because it has only on plane. If it is multi-planer and we use VkBindImagePlaneMemoryInfo to bind memory, the image should be disjoint and the offset is relative to the plane and should be 0 as well.

wenbinc-Bin:
Any other comment? If no, I will submit to community.

xuguangxin (approved):

wenbinc-Bin:
Will submit to community.

PR #436, opened 2021-08-10
Add Deep Link Hyper Encode support

ndomnina:
Hyper Encode uses Intel integrated and discrete graphics on one system to accelerate encoding of a single video stream. Depending on the selected parameters, performance gain on TGL+DG1 compared to TGL: 1.30x-1.60x

Hyper Encode is supported only on Windows and requires D3D11 and oneVPL.

To enable Hyper Encode need to specify:
- Hyper Encode mode (`-dual_gfx on` or `-dual_gfx adaptive`)
- Encoder: h264_qsv or hevc_qsv
- BRC: VBR, CQP or ICQ
- Lowpower mode (`-low_power 1`)
- No specific qsv device

Also requires closed GOP (`-idr_interval 0`, it's used by default)

Depending on the encoding parameters, the following parameters may need to be adjusted:
- `-g` (recommended <= 30 for better performance)
- `-async_depth` (recommended >= 30 for better performance)
- `-extra_hw_frames` (recommended equal to `-async_depth` value to avoid memory allocation issue)

In the cases with fast encoding (`-preset veryfast`) there may be no performance gain due to the fact that the decode is slower than the encode.

Command line example:
`ffmpeg.exe -y -hwaccel qsv -extra_hw_frames 30 -async_depth 30 -c:v h264_qsv -i input.h264 -frames 2000 -async_depth 30 -c:v hevc_qsv -b:v 30000k -maxrate:v 35000k -preset medium -g 30 -low_power 1 -dual_gfx on output.h265`

PR contains two commits:
1) Hyper Encode support. These changes will be ported to cartwheel-ffmpeg
2) Applying patches 0068-0077 (vpl) from cartwheel-ffmpeg, since Hyper Encode is dependent on them

### DIFF
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -41,6 +41,8 @@
 #include "qsv_internal.h"
 #include "qsvenc.h"
 
+#define MFX_IMPL_VIA_MASK(impl) (0x0f00 & (impl))
+
 static const struct {
     mfxU16 profile;
     const char *name;
@@ -298,6 +300,19 @@ static void dump_video_param(AVCodecContext *avctx, QSVEncContext *q,
 
     av_log(avctx, AV_LOG_VERBOSE, "FrameRateExtD: %"PRIu32"; FrameRateExtN: %"PRIu32" \n",
            info->FrameInfo.FrameRateExtD, info->FrameInfo.FrameRateExtN);
+
+#if QSV_HAVE_HE
+    av_log(avctx, AV_LOG_VERBOSE, "HyperEncode: ");
+
+    if (q->hyperEncodeParam.Mode == MFX_HYPERMODE_OFF)
+        av_log(avctx, AV_LOG_VERBOSE, "off");
+    if (q->hyperEncodeParam.Mode == MFX_HYPERMODE_ON)
+        av_log(avctx, AV_LOG_VERBOSE, "on");
+    if (q->hyperEncodeParam.Mode == MFX_HYPERMODE_ADAPTIVE)
+        av_log(avctx, AV_LOG_VERBOSE, "adaptive");
+
+    av_log(avctx, AV_LOG_VERBOSE, "\n");
+#endif
 
 }
 
@@ -825,6 +840,47 @@ static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
         q->extparam_internal[q->nb_extparam_internal++] = (mfxExtBuffer *)&q->extvsi;
     }
 
+    if (q->dual_gfx) {
+#if QSV_HAVE_HE
+        mfxIMPL impl;
+        MFXQueryIMPL(q->session, &impl);
+        if (MFX_IMPL_VIA_MASK(impl) != MFX_IMPL_VIA_D3D11) {
+            av_log(avctx, AV_LOG_ERROR, "Dual GFX mode requires D3D11VA \n");
+        }
+
+        if(q->param.mfx.CodecId != MFX_CODEC_AVC && q->param.mfx.CodecId != MFX_CODEC_HEVC) {
+            av_log(avctx, AV_LOG_WARNING, "Not supported encoder for dual GFX mode. "
+                                          "Supported: h264_qsv and hevc_qsv \n");
+        }
+        if(q->param.mfx.RateControlMethod != MFX_RATECONTROL_VBR &&
+           q->param.mfx.RateControlMethod != MFX_RATECONTROL_CQP &&
+           q->param.mfx.RateControlMethod != MFX_RATECONTROL_ICQ) {
+            av_log(avctx, AV_LOG_WARNING, "Not supported BRC for dual GFX mode. "
+                                          "Supported: VBR, CQP and ICQ \n");
+        }
+        if(q->param.mfx.LowPower != MFX_CODINGOPTION_ON) {
+            av_log(avctx, AV_LOG_WARNING, "Dual GFX mode supports only low-power encoding mode \n");
+        }
+        if((q->param.mfx.CodecId == MFX_CODEC_AVC  && q->param.mfx.IdrInterval != 0) ||
+           (q->param.mfx.CodecId == MFX_CODEC_HEVC && q->param.mfx.IdrInterval != 1)) {
+            av_log(avctx, AV_LOG_WARNING, "Dual GFX mode requires closed GOP (IdrInterval = 0) \n");
+        }
+        if(q->param.mfx.GopPicSize > 30) {
+            av_log(avctx, AV_LOG_WARNING, "For better performance in dual GFX mode GopPicSize must be <= 30 \n");
+        }
+        if(q->param.AsyncDepth < 30) {
+            av_log(avctx, AV_LOG_WARNING, "For better performance in dual GFX mode AsyncDepth must be >= 30 \n");
+        }
+
+        q->hyperEncodeParam.Header.BufferId = MFX_EXTBUFF_HYPER_MODE_PARAM;
+        q->hyperEncodeParam.Header.BufferSz = sizeof(q->hyperEncodeParam);
+        q->hyperEncodeParam.Mode            = q->dual_gfx;
+        q->extparam_internal[q->nb_extparam_internal++] = (mfxExtBuffer *)&q->hyperEncodeParam;
+#else
+        av_log(avctx, AV_LOG_ERROR, "Dual GFX mode is supported only on Windows and requires oneVPL \n");
+#endif
+    }
+
     if (!check_enc_param(avctx,q)) {
         av_log(avctx, AV_LOG_ERROR,
                "some encoding parameters are not supported by the QSV "
@@ -950,7 +1006,14 @@ static int qsv_retrieve_enc_params(AVCodecContext *avctx, QSVEncContext *q)
     };
 #endif
 
-    mfxExtBuffer *ext_buffers[2 + QSV_HAVE_CO2 + QSV_HAVE_CO3 + QSV_HAVE_CO_VPS + QSV_HAVE_EXT_HEVC_TILES];
+#if QSV_HAVE_HE
+    mfxExtHyperModeParam hyperEncodeParam = {
+         .Header.BufferId = MFX_EXTBUFF_HYPER_MODE_PARAM,
+         .Header.BufferSz = sizeof(hyperEncodeParam),
+    };
+#endif
+
+    mfxExtBuffer *ext_buffers[2 + QSV_HAVE_CO2 + QSV_HAVE_CO3 + QSV_HAVE_CO_VPS + QSV_HAVE_EXT_HEVC_TILES + QSV_HAVE_HE];
 
     int need_pps = avctx->codec_id != AV_CODEC_ID_MPEG2VIDEO;
     int ret, ext_buf_num = 0, extradata_offset = 0;
@@ -971,6 +1034,9 @@ static int qsv_retrieve_enc_params(AVCodecContext *avctx, QSVEncContext *q)
 #if QSV_HAVE_EXT_HEVC_TILES
     if (avctx->codec_id == AV_CODEC_ID_HEVC)
         ext_buffers[ext_buf_num++] = (mfxExtBuffer*)&hevc_tile_buf;
+#endif
+#if QSV_HAVE_HE
+    ext_buffers[ext_buf_num++] = (mfxExtBuffer*)&hyperEncodeParam;
 #endif
 
     q->param.ExtParam    = ext_buffers;
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -23,7 +23,7 @@
 
 #include <string.h>
 #include <sys/types.h>
-#include <mfx/mfxvideo.h>
+#include <mfxvideo.h>
 
 #include "libavutil/common.h"
 #include "libavutil/hwcontext.h"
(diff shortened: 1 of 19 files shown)


ndomnina:
@xhaihao, @xuguangxin, @uartie, @wenbinc-Bin, @feiwan1, could you please review this change? I can't add reviewers.

ndomnina:
@xhaihao, @xuguangxin, @uartie, @wenbinc-Bin, @feiwan1, do you have any comments on this change?

libavcodec/qsvenc_h264.c
```diff
@@ -150,6 +150,13 @@ static const AVOption options[] = {
 
     { "repeat_pps", "repeat pps for every frame", OFFSET(qsv.repeat_pps), AV_OPT_TYPE_BOOL, { .i64 = 0 }, 0, 1, VE },
 
+#if QSV_HAVE_HE
+    { "dual_gfx", "Prefer processing on both iGfx and dGfx simultaneously", OFFSET(qsv.dual_gfx), AV_OPT_TYPE_INT, { .i64 = MFX_HYPERMODE_OFF }, MFX_HYPERMODE_OFF, MFX_HYPERMODE_ADAPTIVE, VE, "dual_gfx" },
+    { "off",       NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_HYPERMODE_OFF       },   INT_MIN, INT_MAX, VE, "dual_gfx" },
+    { "on",        NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_HYPERMODE_ON        },   INT_MIN, INT_MAX, VE, "dual_gfx" },
+    { "adaptive",  NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_HYPERMODE_ADAPTIVE  },   INT_MIN, INT_MAX, VE, "dual_gfx" },
+#endif
```
xhaihao: Could you add the options when QSV_HAVE_HE is 0, like what we did in https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/vf_scale_qsv.c#L631-L635? If so, user needn't change the command line when using FFmpeg which is built against MFX_VERSION < 2.0

xhaihao:
What will happen if user specify the device in the command line ? Does HE still work ?

libavcodec/qsvenc_h264.c
ndomnina: I added this, thanks

ndomnina:
@xhaihao, I checked with `-qsv_device` option. If use `-qsv_device 0` and `-dual_gfx on`, then HE works fine. If use `-qsv_device 1` and `-dual_gfx on`, then HE doesn't work, transcoding is performed on the second GPU only. No any errors in this case.

xhaihao:
> @xhaihao, I checked with `-qsv_device` option. If use `-qsv_device 0` and `-dual_gfx on`, then HE works fine. If use `-qsv_device 1` and `-dual_gfx on`, then HE doesn't work, transcoding is performed on the second GPU only. No any errors in this case.

So this is different from the description in the commit log, could you update the commit log ?

ndomnina:
> So this is different from the description in the commit log, could you update the commit log ?

I updated PR/commit messages

ndomnina:
@xhaihao, @xuguangxin, @uartie, @wenbinc-Bin, @feiwan1, @galinart, could you please review this change? Thanks

libavcodec/qsvenc.c
```diff
@@ -823,6 +823,15 @@ static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
         q->extparam_internal[q->nb_extparam_internal++] = (mfxExtBuffer *)&q->extvsi;
     }
 
+#if QSV_HAVE_HE
+    if (q->dual_gfx) {
```
xuguangxin, line 843: Better compare it with MFX_HYPERMODE_OFF.
Users do not need to check oneVPL header to make sure MFX_HYPERMODE_OFF is 0.

libavcodec/qsvenc_h264.c
```diff
@@ -150,6 +150,18 @@ static const AVOption options[] = {
 
     { "repeat_pps", "repeat pps for every frame", OFFSET(qsv.repeat_pps), AV_OPT_TYPE_BOOL, { .i64 = 0 }, 0, 1, VE },
 
+#if QSV_HAVE_HE
+    { "dual_gfx", "Prefer processing on both iGfx and dGfx simultaneously", OFFSET(qsv.dual_gfx), AV_OPT_TYPE_INT, { .i64 = MFX_HYPERMODE_OFF }, MFX_HYPERMODE_OFF, MFX_HYPERMODE_ADAPTIVE, VE, "dual_gfx" },
+    { "off",       NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_HYPERMODE_OFF       },   INT_MIN, INT_MAX, VE, "dual_gfx" },
+    { "on",        NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_HYPERMODE_ON        },   INT_MIN, INT_MAX, VE, "dual_gfx" },
+    { "adaptive",  NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_HYPERMODE_ADAPTIVE  },   INT_MIN, INT_MAX, VE, "dual_gfx" },
```
xuguangxin: could you give some command line in the command line, tell the customer what is adaptive?

libavcodec/qsvenc_hevc.c
```diff
@@ -248,6 +248,18 @@ static const AVOption options[] = {
     { "recovery_point_sei", "Insert recovery point SEI messages",       OFFSET(qsv.recovery_point_sei),      AV_OPT_TYPE_INT, { .i64 = -1 },               -1,          1, VE },
     { "aud", "Insert the Access Unit Delimiter NAL", OFFSET(qsv.aud), AV_OPT_TYPE_INT, { .i64 = 0 }, 0, 1, VE},
 
+#if QSV_HAVE_HE
```
xuguangxin: to reduce the duplication, could we define QSV_HE_OPTIONS like QSV_COMMON_OPTS

libavcodec/qsvenc.c
ndomnina: Thanks, fixed

libavcodec/qsvenc_h264.c
ndomnina: Added the description

libavcodec/qsvenc_hevc.c
ndomnina: Yes, done

galinart:
Please add example of command line to the commit description and add logging into dump_video_param() function

ndomnina:
> Please add example of command line to the commit description and add logging into dump_video_param() function

Added command line example and logging

uartie:
Seems no more reviews.  Thus, please fix rebase conflicts.  If checks pass afterwards, please close this and submit to upstream for community review.

xhaihao:
@uartie This patch depends on oneVPL patchset which is still not merged.

ndomnina:
Performance regression was found with -preset veryfast and Hyper Encode enabled. We working on fix on Hyper Encode side.
After that this patch will be submit to cartwheel-ffmpeg, because oneVPL is only there.

libavcodec/qsvenc.c
```diff
@@ -823,6 +838,47 @@ static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
...
+        if(q->param.mfx.RateControlMethod != MFX_RATECONTROL_VBR &&
+           q->param.mfx.RateControlMethod != MFX_RATECONTROL_CQP &&
+           q->param.mfx.RateControlMethod != MFX_RATECONTROL_ICQ) {
+            av_log(avctx, AV_LOG_WARNING, "Not supported BRC for dual GFX mode. "
+                                          "Supported: VBR, CQP and ICQ \n");
+        }
+        if(q->param.mfx.LowPower != MFX_CODINGOPTION_ON) {
+            av_log(avctx, AV_LOG_WARNING, "Dual GFX mode supports only low-power encoding mode \n");
+        }
+        if((q->param.mfx.CodecId == MFX_CODEC_AVC  && q->param.mfx.IdrInterval != 0) ||
```
galinart, line 864: For AVC IdrInterval !=0 and HEVC IdrInterval !=1? Just to ensure that it is different explicitly?

galinart: Commit message: "Also requires closed GOP (-idr_interval 0, it's used by default)"

libavcodec/qsvenc.c
```diff
@@ -823,6 +838,47 @@ static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
...
+           (q->param.mfx.CodecId == MFX_CODEC_HEVC && q->param.mfx.IdrInterval != 1)) {
+            av_log(avctx, AV_LOG_WARNING, "Dual GFX mode requires closed GOP (IdrInterval = 0) \n");
+        }
+        if(q->param.mfx.GopPicSize > 30) {
+            av_log(avctx, AV_LOG_WARNING, "For better performance in dual GFX mode GopPicSize must be <= 30 \n");
+        }
+        if(q->param.AsyncDepth < 30) {
+            av_log(avctx, AV_LOG_WARNING, "For better performance in dual GFX mode AsyncDepth must be >= 30 \n");
```
galinart, line 872: I believe it should be errors not logs, and return from the function

libavcodec/qsvenc.c
```diff
@@ -823,6 +838,47 @@ static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
...
+        }
+
+        mfxIMPL impl;
+        MFXQueryIMPL(q->session, &impl);
+        if (MFX_IMPL_VIA_MASK(impl) != MFX_IMPL_VIA_D3D11) {
+            av_log(avctx, AV_LOG_WARNING, "Dual GFX mode requires D3D11VA \n");
```
galinart: I believe it is not warning it is mandatory requirement

libavcodec/qsvenc.c
ndomnina: Yes, that's right. For ffmpeg user there is no difference, since ffmpeg set +1 for HEVC internally. But in this place, real values ​​are checked and for different codecs they mean different things. From the specification:

>For H.264, `IdrInterval` specifies IDR-frame interval in terms of I-frames; if `IdrInterval = 0`, then every I-frame is an IDR-frame. If `IdrInterval = 1`, then every other I-frame is an IDR-frame, etc.

>For HEVC, if `IdrInterval = 0`, then only first I-frame is an IDR-frame. If `IdrInterval = 1`, then every I-frame is an IDR-frame. If `IdrInterval = 2`, then every other I-frame is an IDR-frame, etc.

libavcodec/qsvenc.c
ndomnina: These checks are also applied in Adaptive mode, which allows to go into a single GPU fallback if some of the parameters do not meet the conditions. If we will return errors here, then this mode will not work. Do we really need to return errors?

As for "For better performance" conditions, there is no definite answer, which parameters should be specified, since recommendations are different for different command line parameters. I specified 30 here, as we recommend in samples, but with some parameters in ffmpeg, I can achieve good performance with lower values. I.e. <30 can be applicable according to the situation.

galinart: understood, make sense then

libavcodec/qsvenc.c
ndomnina: It is mandatory requirement for VPL in general. I think we can return an error here

ndomnina: Converted to error

libavcodec/qsvenc.c
```diff
@@ -825,6 +840,47 @@ static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
...
+            av_log(avctx, AV_LOG_WARNING, "For better performance in dual GFX mode GopPicSize must be <= 30 \n");
+        }
+        if(q->param.AsyncDepth < 30) {
+            av_log(avctx, AV_LOG_WARNING, "For better performance in dual GFX mode AsyncDepth must be >= 30 \n");
+        }
+
+        mfxIMPL impl;
+        MFXQueryIMPL(q->session, &impl);
+        if (MFX_IMPL_VIA_MASK(impl) != MFX_IMPL_VIA_D3D11) {
+            av_log(avctx, AV_LOG_ERROR, "Dual GFX mode requires D3D11VA \n");
```
galinart: I believe in the case of DX9 it will fail even for ADAPTIVE mode? Right?
Should we get into all of these conditions for VAAPI on Linux if not supported?
Maybe better to check as first condition just after #if QSV_HAVE_HE 844 line and exit if not supported?

ndomnina: > I believe in the case of DX9 it will fail even for ADAPTIVE mode? Right?

Yes, rigth

> Should we get into all of these conditions for VAAPI on Linux if not supported?

I suppose that with VAAPI on Linux we won't get into all of these conditions, because:
```
if (q->dual_gfx) {
#if QSV_HAVE_HE
         ...
#else
        av_log(avctx, AV_LOG_WARNING, "Dual GFX mode is supported only on Windows and requires oneVPL \n");
}
```
and
```
#if defined(_WIN32) || defined(__CYGWIN__)
#define QSV_HAVE_HE     QSV_ONEVPL
#else
#define QSV_HAVE_HE     0
#endif
```

> Maybe better to check as first condition just after #if QSV_HAVE_HE 844 line and exit if not supported?

But even so it probably makes sense (I don't know if this is checked for VPL). I will fix. I think we can also return an error for:
```
av_log(avctx, AV_LOG_WARNING, "Dual GFX mode is supported only on Windows and requires oneVPL \n");
```

ndomnina: Done

xuguangxin:
@noirbee  are we ready for merge this?
thanks

ndomnina:
@xuguangxin, hi. Performance regression with -preset veryfast and Hyper Encode enabled was fixed on Hyper Encode side, but @galinart found a few more problems that need to be fixed before merge (@galinart, please correct if I'm wrong). We are working on it.

galinart:
@ndomnina yes, need to wait a fixes fir other issues. In terms of performance we are good.

PR #440, opened 2021-08-27
libavcodec/dxva2_av1.c: fixed superres_denom param

tong1wu:
fixed superres_denom to solve the problem of the denominator being 0.

### DIFF
--- a/libavcodec/dxva2_av1.c
+++ b/libavcodec/dxva2_av1.c
@@ -73,7 +73,7 @@ static int fill_picture_parameters(const AVCodecContext *avctx, AVDXVAContext *c
     pp->max_height = seq->max_frame_height_minus_1 + 1;
 
     pp->CurrPicTextureIndex = ff_dxva2_get_surface_index(avctx, ctx, h->cur_frame.tf.f);
-    pp->superres_denom      = frame_header->use_superres ? frame_header->coded_denom : AV1_SUPERRES_NUM;
+    pp->superres_denom      = frame_header->use_superres ? frame_header->coded_denom + AV1_SUPERRES_DENOM_MIN : AV1_SUPERRES_NUM;
     pp->bitdepth            = get_bit_depth_from_seq(seq);
     pp->seq_profile         = seq->seq_profile;
 


tong1wu:
@xuguangxin @xhaihao @feiwan1 please help me review this patch and the last patch, thanks.

libavcodec/dxva2_av1.c
```diff
@@ -73,7 +73,7 @@ static int fill_picture_parameters(const AVCodecContext *avctx, AVDXVAContext *c
     pp->max_height = seq->max_frame_height_minus_1 + 1;
 
     pp->CurrPicTextureIndex = ff_dxva2_get_surface_index(avctx, ctx, h->cur_frame.tf.f);
-    pp->superres_denom      = frame_header->use_superres ? frame_header->coded_denom : AV1_SUPERRES_NUM;
+    pp->superres_denom      = frame_header->use_superres ? frame_header->coded_denom + AV1_SUPERRES_DENOM_MIN : AV1_SUPERRES_NUM;
```
xuguangxin, line 76: why we need this change, could you provide some comments in the commit log? the comment better link to some reference.

feiwan1:
Better to change title to "avcodec/dxva2: XXXXX".

libavcodec/dxva2_av1.c
xuguangxin: Defined in spec 5.9.8. When superres is enabled, SuperresDenom equals "coded_denom + SUPERRES_DENOM_MIN" not  coded_denom. 
 
Maybe this is better.

tong1wu: ok, just modified it like that

xuguangxin:
@tong1wu , just for double-check. It's not used by DXVA2 only, D3D11 will use the code too, right?

tong1wu:
@xuguangxin yes basically dxva2 and d3d11 both use the dxva2_av1.c to fill parameters.

xhaihao:
upstream patch: http://ffmpeg.org/pipermail/ffmpeg-devel/2021-September/284920.html

PR #441, opened 2021-08-27
avcodec/av1dec: check if hwaccel is specificed

feiwan1:
Since av1 decoder is only available for hw acceleration. This will
gives out more accurate information if this decoder used but doesn't
specificed a hwaccel.

For example:
ffmpeg -c:v av1 -i INPUT OUTPUT


feiwan1:
@tong1wu helped verified on windows with DXVA2. Result looks good.

PR #443, opened 2021-08-30
qsv: add support for 444 8bit decoding / encoding 

xhaihao:
https://github.com/Intel-Media-SDK/MediaSDK/pull/2768 is needed for testing.


uartie (dismissed):
LGTM

PR #446, opened 2021-09-03
Gaussian Blur Optimization with Vulkan

iqsxw:
This commit adds a powerful and customizable gblur Vulkan filter,
which provides a maximum 127x127 kernel size of Gaussian Filter.
The size could be adjusted by requirements on quality or performance.

Signed-off-by: Wu Jianhua <jianhua.wu@intel.com>


libavfilter/vf_gblur_vulkan.c
```diff
@@ -0,0 +1,472 @@
...
+    av_frame_free(&tmp);
+    av_frame_free(&out);
+
+    return err;
+}
+
+#define OFFSET(x) offsetof(GBlurVulkanContext, x)
+#define FLAGS (AV_OPT_FLAG_FILTERING_PARAM | AV_OPT_FLAG_VIDEO_PARAM)
+static const AVOption gblur_vulkan_options[] = {
+    { "sigma", "Set sigma", OFFSET(sigma), AV_OPT_TYPE_FLOAT, {.dbl = 0.5}, 0.001, 1024.0, FLAGS },
```
xuguangxin: why it's 0.001 not 0?

libavfilter/vf_gblur_vulkan.c
```diff
@@ -0,0 +1,472 @@
...
+    { "size", "Set kernel Size(If not set, generated automatically for better quality)", OFFSET(size), AV_OPT_TYPE_INT, {.i64 = 19}, 1, GBLUR_MAX_KERNEL_SIZE, FLAGS },
```
xuguangxin: why we need kernel size? cpu version not needed?

libavfilter/vf_gblur_vulkan.c
```diff
@@ -0,0 +1,472 @@
...
+        ff_vk_update_descriptor_set(ctx, s->pl_ver, 1);
+    }
+
+    RET(ff_vk_create_exec_ctx(ctx, &s->exec));
+
+    s->initialized = 1;
+    av_free(kernel_def);
+
+    return 0;
+fail:
```
xuguangxin, line 301: do we need free kernel_def?

iqsxw: Sure. If not we may have a memory leak here. The memory chunk for kernel_def was allocated in the function, av_asprintf.

libavfilter/vf_gblur_vulkan.c
iqsxw: The reason here is that using zero as the variance does nothing. As for 0.001, it's to avoid the user from entering a too-small value. Otherwise, the subsequent operation's a waste of resources and time.

libavfilter/vf_gblur_vulkan.c
iqsxw: The CPU version uses the whole image extent as its kernel size, which is unnecessary to be specified. This is more about the difference between the execution manners of CPU and GPU.

libavfilter/vf_gblur_vulkan.c
xuguangxin: If we use 0, it will fail on GPU and success on CPU.

iqsxw:
I'll close this pull request, for, the patch has been merged into the upstream,

PR #448, opened 2021-09-09
libavcodec/hevc_mp4toannexb_bsf: ignore extra data if possible

xhaihao:
It is possible that an IRAP frame in in-AVPacket has VPS, SPS and PPS,
and these headere should take effect in out-AVPacket, so we should not
prepend extradata to IRAP frame in this case, otherwise an IRAP frame in
out-AVPcket might have 2 VPS/SPS/PPS if extradata has VPS, SPS and PPS,
the second VPS/SPS/PPS will take effect

### DIFF
--- a/libavcodec/hevc_mp4toannexb_bsf.c
+++ b/libavcodec/hevc_mp4toannexb_bsf.c
@@ -121,7 +121,7 @@ static int hevc_mp4toannexb_filter(AVBSFContext *ctx, AVPacket *out)
     HEVCBSFContext *s = ctx->priv_data;
     AVPacket *in;
     GetByteContext gb;
-
+    int has_sps = 0, has_pps = 0;
     int got_irap = 0;
     int i, ret = 0;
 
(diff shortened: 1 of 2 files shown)


libavcodec/hevc_mp4toannexb_bsf.c
```diff
@@ -155,10 +155,14 @@ static int hevc_mp4toannexb_filter(AVBSFContext *ctx, AVPacket *out)
         }
 
         nalu_type = (bytestream2_peek_byte(&gb) >> 1) & 0x3f;
+        has_vps = (has_vps || nalu_type == HEVC_NAL_VPS);
```
xuguangxin: vps is optional. most bitstream does not have it.

libavcodec/hevc_mp4toannexb_bsf.c
```diff
@@ -155,10 +155,14 @@ static int hevc_mp4toannexb_filter(AVBSFContext *ctx, AVPacket *out)
...
+        has_sps = (has_sps || nalu_type == HEVC_NAL_SPS);
+        has_pps = (has_pps || nalu_type == HEVC_NAL_PPS);
 
         /* prepend extradata to IRAP frames */
         is_irap       = nalu_type >= 16 && nalu_type <= 23;
-        add_extradata = is_irap && !got_irap;
+        /* ignore the extradata if IRAP frame has vps, sps and pps */
+        add_extradata = is_irap && !got_irap && (!has_vps || !has_sps || !has_pps);
```
xuguangxin: the comment may not match the code. The code means, once we see sps, pps in the stream, we will not add extra data.

xhaihao: The extradata is ignored if a IRAP frame already has all of vps, sps and pps, so we won't prepend the data to out AVPacket. I think it is the same of what you said.

libavcodec/hevc_mp4toannexb_bsf.c
xhaihao: I am aware that vps is optional. This patch is only used to fix some cases, not all cases.

xhaihao: but I may remove has_vps in a patch version.

xhaihao: removed has_vps in the new version

xuguangxin (dismissed):

xhaihao:
Update fate ref for fate-hevc-bsf-mp4toannexb case

xhaihao:
upstream patch: http://ffmpeg.org/pipermail/ffmpeg-devel/2021-September/285031.html

PR #454, opened 2021-09-16
libavcodec/vaapi_encode: Change the way to call async to increase

wenbinc-Bin:
performance

Fix: #7706. After commit 5fdcf85bbffe7451c2, vaapi encoder's performance
decrease. The reason is that vaRenderPicture() and vaSyncSurface() are
called at the same time (vaRenderPicture() always followed by a
vaSyncSurface()). Now I changed them to be called in a
asynchronous way, which will make better use of hardware.

### DIFF
--- a/libavcodec/vaapi_encode.c
+++ b/libavcodec/vaapi_encode.c
@@ -951,8 +951,10 @@ static int vaapi_encode_pick_next(AVCodecContext *avctx,
     if (!pic && ctx->end_of_stream) {
         --b_counter;
         pic = ctx->pic_end;
-        if (pic->encode_issued)
+        if (pic->encode_complete)
             return AVERROR_EOF;
+        else if (pic->encode_issued)
+            return AVERROR(EAGAIN);
     }
 
     if (!pic) {
@@ -1177,19 +1179,30 @@ int ff_vaapi_encode_receive_packet(AVCodecContext *avctx, AVPacket *pkt)
             return AVERROR(EAGAIN);
     }
 
-    pic = NULL;
-    err = vaapi_encode_pick_next(avctx, &pic);
-    if (err < 0)
-        return err;
-    av_assert0(pic);
+    while (av_fifo_size(ctx->encode_fifo) <= MAX_PICTURE_REFERENCES * sizeof(VAAPIEncodePicture *)) {
+        pic = NULL;
+        err = vaapi_encode_pick_next(avctx, &pic);
+        if (err < 0)
+            break;
+        av_assert0(pic);
 
-    pic->encode_order = ctx->encode_order++;
+        pic->encode_order = ctx->encode_order +
+                            (av_fifo_size(ctx->encode_fifo) / sizeof(VAAPIEncodePicture *));
 
-    err = vaapi_encode_issue(avctx, pic);
-    if (err < 0) {
-        av_log(avctx, AV_LOG_ERROR, "Encode failed: %d.\n", err);
-        return err;
+        err = vaapi_encode_issue(avctx, pic);
+        if (err < 0) {
+            av_log(avctx, AV_LOG_ERROR, "Encode failed: %d.\n", err);
+            return err;
+        }
+
+        av_fifo_generic_write(ctx->encode_fifo, &pic, sizeof(pic), NULL);
     }
+
+    if (!av_fifo_size(ctx->encode_fifo))
+        return err;
+
+    av_fifo_generic_read(ctx->encode_fifo, &pic, sizeof(pic), NULL);
+    ctx->encode_order = pic->encode_order + 1;
 
     err = vaapi_encode_output(avctx, pic, pkt);
     if (err < 0) {
@@ -2520,6 +2533,11 @@ av_cold int ff_vaapi_encode_init(AVCodecContext *avctx)
         }
     }
 
+    ctx->encode_fifo = av_fifo_alloc((MAX_PICTURE_REFERENCES + 1) *
+                                      sizeof(VAAPIEncodePicture *));
+    if (!ctx->encode_fifo)
+        return AVERROR(ENOMEM);
+
     return 0;
 
 fail:
@@ -2552,6 +2570,7 @@ av_cold int ff_vaapi_encode_close(AVCodecContext *avctx)
 
     av_freep(&ctx->codec_sequence_params);
     av_freep(&ctx->codec_picture_params);
+    av_fifo_freep(&ctx->encode_fifo);
 
     av_buffer_unref(&ctx->recon_frames_ref);
     av_buffer_unref(&ctx->input_frames_ref);
--- a/libavcodec/vaapi_encode.c
+++ b/libavcodec/vaapi_encode.c
@@ -1158,7 +1158,8 @@ static int vaapi_encode_send_frame(AVCodecContext *avctx, AVFrame *frame)
         if (ctx->input_order == ctx->decode_delay)
             ctx->dts_pts_diff = pic->pts - ctx->first_pts;
         if (ctx->output_delay > 0)
-            ctx->ts_ring[ctx->input_order % (3 * ctx->output_delay)] = pic->pts;
+            ctx->ts_ring[ctx->input_order %
+                        (3 * ctx->output_delay + ctx->async_depth)] = pic->pts;
 
         pic->display_order = ctx->input_order;
         ++ctx->input_order;
@@ -1212,7 +1213,8 @@ int ff_vaapi_encode_receive_packet(AVCodecContext *avctx, AVPacket *pkt)
             return AVERROR(EAGAIN);
     }
 
-    while (av_fifo_size(ctx->encode_fifo) <= MAX_PICTURE_REFERENCES * sizeof(VAAPIEncodePicture *)) {
+    while (av_fifo_size(ctx->encode_fifo) <
+            MAX_ASYNC_DEPTH * sizeof(VAAPIEncodePicture *)) {
         pic = NULL;
         err = vaapi_encode_pick_next(avctx, &pic);
         if (err < 0)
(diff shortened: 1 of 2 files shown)


wenbinc-Bin:
On my side, Transcoding 1080p h264 on i9-9900k increase from 240fps to 270fps.

libavcodec/vaapi_encode.c
```diff
@@ -1177,20 +1179,34 @@ int ff_vaapi_encode_receive_packet(AVCodecContext *avctx, AVPacket *pkt)
             return AVERROR(EAGAIN);
     }
 
-    pic = NULL;
-    err = vaapi_encode_pick_next(avctx, &pic);
-    if (err < 0)
-        return err;
-    av_assert0(pic);
+    if (av_fifo_size(ctx->encode_fifo) == 0) {
```
feiwan1: Any  possible to remove this check? Or check fifo size inside while with "if (av_fifo_size(ctx->encode_fifo) == MAX_ALLOC_SIZE) break;". With your logic, if fifo size is 0, and the next pic will write to fifo, and then get it out at once, this meets the case in your commit log. If change like this, you may consider if need to enlarge fifo size in init function.

wenbinc-Bin: Thanks, I try to improve code like you said.

wenbinc-Bin: Done, please check.

libavcodec/vaapi_encode.c
```diff
@@ -1177,20 +1179,31 @@ int ff_vaapi_encode_receive_packet(AVCodecContext *avctx, AVPacket *pkt)
...
+        }
+
+        av_fifo_generic_write(ctx->encode_fifo, &pic, sizeof(pic), NULL);
     }
 
+    if (av_fifo_size(ctx->encode_fifo) == 0)
+        return err;
+
+    av_fifo_generic_read(ctx->encode_fifo, &pic, sizeof(pic), NULL);
+    ctx->encode_order = pic->encode_order+1;
```
feiwan1: "encoder_order+1" -> "encoder_order + 1".

libavcodec/vaapi_encode.c
```diff
@@ -2520,6 +2533,10 @@ av_cold int ff_vaapi_encode_init(AVCodecContext *avctx)
         }
     }
 
+    ctx->encode_fifo = av_fifo_alloc((MAX_PICTURE_REFERENCES+1)* sizeof(VAAPIEncodePicture *));
```
feiwan1: "MAX_PICTURE_REFERENCES+1)*" -> "MAX_PICTURE_REFERENCES + 1) *". You can use tools/patchcheck to scan your patch, it can detect this kind of style issue. Others looks good to me.

libavcodec/vaapi_encode.c
```diff
@@ -1177,20 +1179,31 @@ int ff_vaapi_encode_receive_packet(AVCodecContext *avctx, AVPacket *pkt)
             return AVERROR(EAGAIN);
     }
 
-    pic = NULL;
-    err = vaapi_encode_pick_next(avctx, &pic);
-    if (err < 0)
-        return err;
-    av_assert0(pic);
+    while (av_fifo_size(ctx->encode_fifo) <= MAX_PICTURE_REFERENCES*sizeof(VAAPIEncodePicture *)) {
```
feiwan1: " MAX_PICTURE_REFERENCES*sizeof" -> " MAX_PICTURE_REFERENCES * sizeof"

libavcodec/vaapi_encode.c
wenbinc-Bin: done

libavcodec/vaapi_encode.c
wenbinc-Bin: done

libavcodec/vaapi_encode.c
wenbinc-Bin: done

wenbinc-Bin:
Any other comments?

wenbinc-Bin:
Will submit to upstream if there is no other comment.

xuguangxin (commented):
Will this introduce frame delay for IPPPPP pattern?

xuguangxin:
Is it possible to use https://github.com/intel/libva/blob/master/va/va.h#L4193 to query buffer status?
the logical maybe:
1. vaapi_encode_issue
2. push to fifo
3. if fifo is full, vaSyncBuffer with infinitely.  pop and push it to downstream. 
4. vaSyncBuffer the entire fifo with timeout 0. stop if we find a VA_STATUS_ERROR_TIMEDOUT

wenbinc-Bin:
@xuguangxin, Ok I will try this to improve performance.

libavcodec/vaapi_encode_vp9.c
```diff
@@ -260,6 +260,13 @@ static const AVOption vaapi_encode_vp9_options[] = {
       OFFSET(loop_filter_level), AV_OPT_TYPE_INT, { .i64 = 16 }, 0, 63, FLAGS },
     { "loop_filter_sharpness", "Loop filter sharpness",
       OFFSET(loop_filter_sharpness), AV_OPT_TYPE_INT, { .i64 = 4 }, 0, 15, FLAGS },
+
+#if VA_CHECK_VERSION(1, 9, 0)
```
xuguangxin: could we add this to VAAPI_ENCODE_COMMON_OPTIONS

libavcodec/vaapi_encode.c
```diff
@@ -2520,6 +2550,13 @@ av_cold int ff_vaapi_encode_init(AVCodecContext *avctx)
         }
     }
 
+    ctx->encode_fifo = av_fifo_alloc((ctx->async_depth > MAX_PICTURE_REFERENCES ?
```
xuguangxin: 1. If we have this, How to handle low delay encode? 
2. could we adjust async_depth in one place?

libavcodec/vaapi_encode.c
```diff
@@ -1177,19 +1180,46 @@ int ff_vaapi_encode_receive_packet(AVCodecContext *avctx, AVPacket *pkt)
...
+            return err;
+        }
+
+        av_fifo_generic_write(ctx->encode_fifo, &pic, sizeof(pic), NULL);
+    }
+
+    if (!av_fifo_size(ctx->encode_fifo))
         return err;
+
+#if VA_CHECK_VERSION(1, 9, 0)
```
xuguangxin: if < 1.9.0, we can use vaQuerySurface

libavcodec/vaapi_encode.c
```diff
@@ -1177,19 +1180,46 @@ int ff_vaapi_encode_receive_packet(AVCodecContext *avctx, AVPacket *pkt)
...
+    if (av_fifo_size(ctx->encode_fifo) < ctx->async_depth * sizeof(VAAPIEncodePicture *) &&
+        !ctx->end_of_stream) {
+        av_fifo_generic_peek(ctx->encode_fifo, &pic, sizeof(pic), NULL);
+        err = vaSyncBuffer(ctx->hwctx->display, pic->output_buffer, 0);
```
xuguangxin: we'd better output all ready buffers. not only the current buffer, it will reduce the delay

libavcodec/vaapi_encode_vp9.c
wenbinc-Bin: I tried but #if cannot be added into macro definition.

xuguangxin: you can use vaQuerySurface for < 1.9

libavcodec/vaapi_encode.c
wenbinc-Bin: 1.The delay is caused by async_depth not by the size of encode_fifo.
I create encode_fifo with at least size 3 to make sure we get high performance when encode b frames. This will not cause delay.
2. if I use vaQuerySurface when < 1.9.0, I can move async_depth option to VAAPI_ENCODE_COMMON_OPTIONS, then we can adjust this variable through command line option.

libavcodec/vaapi_encode.c
wenbinc-Bin: Thanks for the information, I will try.

libavcodec/vaapi_encode.c
```diff
@@ -150,12 +150,21 @@ static int vaapi_encode_wait(AVCodecContext *avctx,
            "(input surface %#x).\n", pic->display_order,
            pic->encode_order, pic->input_surface);
 
+#if VA_CHECK_VERSION(1, 9, 0)
```
xuguangxin, line 154: if vaSyncBuffer returns unimplemented we may need fallback to vaSyncSurface.

libavcodec/vaapi_encode.h
```diff
@@ -455,7 +462,10 @@ int ff_vaapi_encode_close(AVCodecContext *avctx);
     { "b_depth", \
       "Maximum B-frame reference depth", \
       OFFSET(common.desired_b_depth), AV_OPT_TYPE_INT, \
-      { .i64 = 1 }, 1, INT_MAX, FLAGS }
+      { .i64 = 1 }, 1, INT_MAX, FLAGS }, \
+    { "async_depth", "Maximum processing parallelism", \
```
xuguangxin: better say, increase this to improve single channel performance

libavcodec/vaapi_encode.c
```diff
@@ -1177,20 +1189,60 @@ int ff_vaapi_encode_receive_packet(AVCodecContext *avctx, AVPacket *pkt)
...
+        err = vaSyncBuffer(ctx->hwctx->display, pic->output_buffer, 0);
+        if (err == VA_STATUS_ERROR_TIMEDOUT) {
+            return AVERROR(EAGAIN);
+        } else if (err != VA_STATUS_SUCCESS){
+            av_log(avctx, AV_LOG_ERROR, "Failed to sync buffers: "
+                   "%d (%s).\n", err, vaErrorStr(err));
+            return AVERROR(EIO);
+        }
+#elif
```
xuguangxin: we can add a timeout to vaapi_encode_wait. and call vaapi_encode_wait(..., timeout = 0) here

libavcodec/vaapi_encode_vp9.c
wenbinc-Bin: done

libavcodec/vaapi_encode.c
wenbinc-Bin: Oh, that is why pre-check fail in i965 test.
Thanks

libavcodec/vaapi_encode.c
wenbinc-Bin: Good idea, will change.
Thanks

libavcodec/vaapi_encode.h
wenbinc-Bin: done

xuguangxin (approved):

wenbinc-Bin:
submit to https://patchwork.ffmpeg.org/project/ffmpeg/patch/20211027085705.4114165-1-wenbin.chen@intel.com/

PR #455, opened 2021-09-17
take HW acceleration method into account when selecting decoder for input stream

xhaihao:


xhaihao:
Any comment for this PR? We may simplify the command when using av1 vaapi decoder and all qsv decoders after applying this patchset.

wenbinc-Bin:
> Any comment for this PR? We may simplify the command when using av1 vaapi decoder and all qsv decoders after applying this patchset.

Yes, I tried. Now I don't need to specify "-c:v h264_qsv" to use qsv decoder, only "-hwaccel qsv" is enough. Thanks.
BTW, Code looks good to me. Thanks

xhaihao:
upstream patch: https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=4924

PR #457, opened 2021-09-18
qsvenc: re-initialize MFX session if needed

xhaihao:
Example:
$> ffmpeg -stream_loop -1 -hwaccel qsv -c:v h264_qsv -i input.mp4 \
-autoscale 0 -c:v h264_qsv -f null -

...
[h264_qsv @ 0x55f91bc4b240] Error submitting the frame for encoding.
Video encoding failed

The above issue disappears after applying this patch

### DIFF
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -1546,6 +1546,14 @@ int ff_qsv_encode(AVCodecContext *avctx, QSVEncContext *q,
 {
     int ret;
 
+    if (frame && avctx->hw_frames_ctx && frame->hw_frames_ctx &&
+        avctx->hw_frames_ctx->data != frame->hw_frames_ctx->data) {
+        avctx->codec->close(avctx);
+        av_buffer_unref(&avctx->hw_frames_ctx);
+        avctx->hw_frames_ctx = av_buffer_ref(frame->hw_frames_ctx);
+        avctx->codec->init(avctx);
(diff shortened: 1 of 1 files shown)


wenbinc-Bin:
@xhaihao The another cause is that https://github.com/wenbinc-Bin/ffmpeg-ci/blob/51f1194edae2020ec99b816bd045a29db0e469f8/libavcodec/qsvdec.c#L818 flush function in qsvdec set orig_pix_fmt to none, that causes reinitialization of decoder. Do we need to remove this code? After applying your patch, the encoder will reinit when one loop end, so at this border the output video may not encoded as configuration, for example gop_size.

xhaihao:
> @xhaihao The another cause is that https://github.com/wenbinc-Bin/ffmpeg-ci/blob/51f1194edae2020ec99b816bd045a29db0e469f8/libavcodec/qsvdec.c#L818 flush function in qsvdec set orig_pix_fmt to none, that causes reinitialization of decoder. Do we need to remove this code?

I'm aware of this code and my initial patch was to remove this code.  but hwaccel_uninit / hwaccel_init for other HW decoders are called too after avcodec_flush_buffers is called, so I didn't remove this code in this PR.

> After applying your patch, the encoder will reinit when one loop end, so at this border the output video may not encoded as configuration, for example gop_size.

Thanks for raising this issue, I will look into it.

PR #463, opened 2021-10-08
Onevpl upstream

xhaihao:
run pre-check for new patchset version


libavutil/hwcontext_qsv.c
```diff
@@ -525,13 +550,247 @@ static mfxStatus frame_get_hdl(mfxHDL pthis, mfxMemId mid, mfxHDL *hdl)
...
+               "MFX_ACCEL_MODE_VIA_D3D9 property: %d.\n", sts);
+        goto fail;
+    }
+
+    if (MFX_HANDLE_D3D9_DEVICE_MANAGER == handle_type ||
+        MFX_HANDLE_D3D11_DEVICE == handle_type) {
+        uint32_t idx = 0;
+
+        if (child_device)
+            idx = atoi(child_device);
```
galinart: might be better to add MFXSetConfigFilterProperty only if child_device != NULL

Otherwise not clear why we request idx = 0 only if not in "if (child_device)" condition. 
Potentially it could be any index, system dependent I believe. e.g. Nvidia GPUs + Intel GPUs in the system. Hard to say that Intel will get idx = 0 in this case.

xhaihao: Thanks for the comment, I'll update it.

libavutil/hwcontext_dxva2.c
```diff
@@ -570,6 +571,13 @@ static int dxva2_device_create(AVHWDeviceContext *ctx, const char *device,
         return AVERROR_UNKNOWN;
     }
 
+    if (device) {
+        hwctx->adapter = av_strdup(adapter);
```
galinart: const char* device - string  
UINT adapter  - integer

I believe it should be since strdup has char input parameter:

```
 if (device) {
        hwctx->adapter = av_strdup(device);
```

galinart: static int dxva2_device_create9ex(AVHWDeviceContext *ctx, UINT adapter)
static int dxva2_device_create9ex(AVHWDeviceContext *ctx, UINT adapter)
static int dxva2_device_create9(AVHWDeviceContext *ctx, UINT adapter)

static int dxva2_device_create(AVHWDeviceContext *ctx, const char *device,
                               AVDictionary *opts, int flags)

typedef struct AVDXVA2DeviceContext {
    IDirect3DDeviceManager9 *devmgr;
    /**
     * The string for the used adapter
     */
    char *adapter;
} AVDXVA2DeviceContext;

Slightly confusing to have the same "adapter" name but different types. I believe we can change that since all functions are static, no impact on interface and headers

libavcodec/qsv_internal.h
```diff
@@ -90,6 +90,7 @@ typedef struct QSVSession {
 #ifdef AVCODEC_QSV_LINUX_SESSION_HANDLE
     AVBufferRef *va_device_ref;
     AVHWDeviceContext *va_device_ctx;
+    void *loader;
```
galinart:
```
typedef struct QSVSession {
    mfxSession session;
#ifdef AVCODEC_QSV_LINUX_SESSION_HANDLE
    AVBufferRef *va_device_ref;
    AVHWDeviceContext *va_device_ctx;
    void *loader;
#endif
} QSVSession;
```

Should be:

```
typedef struct QSVSession {
    mfxSession session;
    void *loader;
#ifdef AVCODEC_QSV_LINUX_SESSION_HANDLE
    AVBufferRef *va_device_ref;
    AVHWDeviceContext *va_device_ctx;
#endif
} QSVSession;
```

Otherwise on Windows:

```
CC      libavcodec/qsv.o
C:/msys64/home/Artem/FFMPEG-new-session/libavcodec/qsv.c: In function 'ff_qsv_init_internal_session':
C:/msys64/home/Artem/FFMPEG-new-session/libavcodec/qsv.c:566:41: error: 'QSVSession' has no member named 'loader'
  566 |                                      &qs->loader);
      |                                         ^~
C:/msys64/home/Artem/FFMPEG-new-session/libavcodec/qsv.c: In function 'ff_qsv_close_internal_session':
C:/msys64/home/Artem/FFMPEG-new-session/libavcodec/qsv.c:992:21: error: 'QSVSession' has no member named 'loader'
  992 |         MFXUnload(qs->loader);
      |                     ^~
C:/msys64/home/Artem/FFMPEG-new-session/libavcodec/qsv.c:994:11: error: 'QSVSession' has no member named 'loader'
  994 |         qs->loader = NULL;
      |           ^~
make: *** [/home/Artem/F
```

libavutil/hwcontext_qsv.c
```diff
@@ -537,8 +550,241 @@ static mfxStatus frame_get_hdl(mfxHDL pthis, mfxMemId mid, mfxHDL *hdl)
...
+               "property: %d.\n", sts);
+        goto fail;
+    }
+
+    impl_value.Type = MFX_VARIANT_TYPE_U32;
+
+    if (MFX_HANDLE_VA_DISPLAY == handle_type)
+        impl_value.Data.U32 = MFX_ACCEL_MODE_VIA_VAAPI;
+    else if (MFX_HANDLE_D3D9_DEVICE_MANAGER == handle_type)
+        impl_value.Data.U32 = MFX_ACCEL_MODE_VIA_D3D9;
```
galinart, line 618: Think we need to clarify expected behavior for oneVPL in case of MFX_ACCEL_MODE_VIA_D3D9 type

libavutil/hwcontext_dxva2.c
xhaihao: Yes, it is strdup(device) for both dxva2 & d3d11va. 

I named it 'device' initially, but AVD3D11VADeviceContext has already have a member of 'device', and I wanted to use the same name in both AVD3D11VADeviceContext & AVDXVA2DeviceContext.  I will change it to device_name in the new version.

libavcodec/qsv_internal.h
xhaihao: Thanks for catching this issue.

libavutil/hwcontext_vaapi.h
```diff
@@ -78,6 +78,10 @@ typedef struct AVVAAPIDeviceContext {
      * operations using VAAPI with the same VADisplay.
      */
     unsigned int driver_quirks;
+    /**
+     * The string for the used device
+     */
+    char *device;
```
galinart: Just for consistency have the same name for all device types? 
https://github.com/intel-media-ci/ffmpeg/pull/463/commits/d7072fdc00e5628849e3dfff7656ab173acbf130#diff-420c00076aa968374eb6be9ba299e1a8c33bd7a04b6dc9a60b8fdbd78948d4a0R44

xhaihao: Updated, thanks!

libavutil/hwcontext_qsv.c
xhaihao: According to https://spec.oneapi.io/versions/latest/elements/oneVPL/source/programming_guide/VPL_prg_hw.html?highlight=dxva2#work-with-microsoft-directx-applications, oneVPL API may supports D3D9, but the underlying runtime might not support it.

maximd33i:
@xhaihao let's have it shared with ffmpeg-devel ML

galinart:
@xhaihao please update patchwork patchset

PR #464, opened 2021-10-21
lavc/av1dec: drop redundant frame

feiwan1:
For some non-standard clip which contains multiple display frame
data in a ivf packet, drop the redundant frame can make the decode
continuously with some warning message shown. Otherwise decode will
interrupt with unexpected error.


feiwan1:
@xhaihao  @xuguangxin @wenbinc-Bin , any comments on this patch? I am ready to submit it to community. Thanks.

wenbinc-Bin:
I tried on the stream with wrong TU. It works.
code LGTM

PR #486, opened 2021-12-27
Fix handle malloc

wenbinc-Bin:


wenbinc-Bin:
@xhaihao I change the patch like this. Is it ok? Submit it here first to discuss.

PR #495, opened 2022-01-18
Enable libtorch

Semmer2:
Fixed some issues and rebased on latest upstream code.
Current code has been verified on CPU with BasicVSR model.
Now FFmpeg Libtorch can take ambiguous input size.

### DIFF
--- a/libavfilter/dnn/dnn_backend_torch.cpp
+++ b/libavfilter/dnn/dnn_backend_torch.cpp
@@ -223,6 +223,14 @@ DNNModel *ff_dnn_load_model_th(const char *model_filename, DNNFunctionType func_
     c10::Device device = c10::Device(ctx->options.device_name);
     if (device.is_cpu()) {
         ctx->options.device_type = torch::kCPU;
+    } else if (device.is_xpu()) {
+        if (!at::hasXPU()) {
+            av_log(ctx, AV_LOG_ERROR, "No XPU device found\n");
+            return NULL;
+        }
+        at::detail::getXPUHooks().initXPU();
+        at::deviceExclusiveCheck();
+        ctx->options.device_type = torch::kXPU;
     } else {
         av_log(ctx, AV_LOG_ERROR, "Not supported device:\"%s\"\n", ctx->options.device_name);
         goto fail;
@@ -335,6 +343,8 @@ static int fill_model_input_th(THModel *th_model, THRequestItem *request)
     }
     *infer_request->input_tensor = torch::from_blob(input.data, {1, 1, 3, input.height, input.width},
                                                     torch::kFloat32);
+    if (infer_request->input_tensor->device() != ctx->options.device_type)
+            *infer_request->input_tensor = infer_request->input_tensor->to(ctx->options.device_type);
     return 0;
 
 err:
@@ -396,6 +406,9 @@ static void infer_completion_callback(void *args) {
     switch (th_model->model->func_type) {
     case DFT_PROCESS_FRAME:
         if (task->do_ioproc) {
+            //post process can only deal with CPU memory.
+            if (output->device() != torch::kCPU)
+                *output = output->to(torch::kCPU);
             outputs.data = output->data_ptr();
             if (th_model->model->frame_post_proc != NULL) {
                 th_model->model->frame_post_proc(task->out_frame, &outputs, th_model->model->filter_ctx);
--- a/libavfilter/dnn/dnn_backend_torch.cpp
+++ b/libavfilter/dnn/dnn_backend_torch.cpp
@@ -1,0 +1,567 @@
+/*
+ * Copyright (c) 2022
+ *
+ * This file is part of FFmpeg.
+ *
+ * FFmpeg is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * FFmpeg is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with FFmpeg; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+ */
+
+/**
+ * @file
+ * DNN Torch backend implementation.
+ */
+
+#include <torch/torch.h>
+#include <torch/script.h>
+#include "dnn_backend_torch.h"
+
+extern "C" {
+#include "dnn_io_proc.h"
+#include "../internal.h"
+#include "dnn_backend_common.h"
+#include "libavutil/opt.h"
+#include "queue.h"
+#include "safe_queue.h"
+}
+
+typedef struct THOptions{
+    char *device_name;
+    c10::DeviceType device_type;
+} THOptions;
+
+typedef struct THContext {
+    const AVClass *c_class;
+    THOptions options;
+} THContext;
+
+typedef struct THModel {
+    THContext ctx;
+    DNNModel *model;
+    torch::jit::Module jit_model;
+    SafeQueue *request_queue;
+    Queue *task_queue;
+    Queue *lltask_queue;
+} THModel;
+
+typedef struct THInferRequest {
+    torch::Tensor *output;
+    torch::Tensor *input_tensor;
+} THInferRequest;
+
+typedef struct THRequestItem {
+    THInferRequest *infer_request;
+    LastLevelTaskItem *lltask;
+    DNNAsyncExecModule exec_module;
+} THRequestItem;
+
+
+#define OFFSET(x) offsetof(THContext, x)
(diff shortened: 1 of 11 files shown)


Semmer2:
@wenbinc-Bin hi Wenbin, would you please take a look at my code? thanks a lot :)

libavfilter/dnn_interface.h
```diff
@@ -59,9 +53,9 @@ typedef enum {
 typedef struct DNNData{
     void *data;
     int width, height, channels;
-    // dt and order together decide the color format
+    // dt and format together decide the color format
+    enum AVPixelFormat format;
```
wenbinc-Bin: I think you'd better put "format" under "dt". It helps to keep compatibility if user uses offset to access structure.

libswscale/swscale_unscaled.c
```diff
@@ -1095,9 +1090,24 @@ static int planarRgbToRgbWrapper(SwsContext *c, const uint8_t *src[],
         break;
 
     case AV_PIX_FMT_RGB24:
-        gbr24ptopacked24(src201, stride201,
-                         dst[0] + srcSliceY * dstStride[0], dstStride[0],
-                         srcSliceH, c->srcW);
+        switch (c->srcFormat) {
+        case AV_PIX_FMT_RGBP:
```
wenbinc-Bin: You'd better add RGBP support or add srcFormat check to other dstFormat, otherwise it will call function with wrong parameters.

libswscale/swscale_unscaled.c
```diff
@@ -1188,8 +1202,15 @@ static int rgbToPlanarRgbWrapper(SwsContext *c, const uint8_t *src[],
 
     switch (c->srcFormat) {
     case AV_PIX_FMT_RGB24:
-        packedtogbr24p((const uint8_t *) src[0], srcStride[0], dst201,
-                       stride201, srcSliceH, alpha_first, 3, c->srcW);
+        switch (c->dstFormat) {
+        case AV_PIX_FMT_RGBP:
```
wenbinc-Bin: Same as above, you'd better consider RGBP on other srcFormat branch.

wenbinc-Bin (commented):
I checked your first three patches and I am checking your last one. Because the last patch is a little long, it will take me some time to reply.
Thanks
Wenbin

libavfilter/dnn/dnn_io_proc.c
```diff
@@ -35,6 +41,18 @@ DNNReturnType ff_proc_from_dnn_to_frame(AVFrame *frame, DNNData *output, void *l
         avpriv_report_missing_feature(log_ctx, "data type rather than DNN_FLOAT");
         return DNN_ERROR;
     }
+    if (output->format == AV_PIX_FMT_RGBP) {
+        tmp_data = malloc(frame_size * 3 * sizeof(uint8_t));
```
wenbinc-Bin: tmp_data is not freed.

libavfilter/dnn/dnn_backend_torch.cpp
```diff
@@ -0,0 +1,497 @@
...
+err:
+    th_free_request(infer_request);
+
+    if (ff_safe_queue_push_back(th_model->request_queue, request) < 0) {
+        destroy_request_item(&request);
+        av_log(&th_model->ctx, AV_LOG_ERROR, "Unable to push back request_queue when failed to start inference.\n");
+    }
+}
+
+DNNReturnType execute_model_th(THRequestItem *request, Queue *lltask_queue)
```
wenbinc-Bin: You maybe forget to define this function as static.

libavfilter/dnn/dnn_backend_torch.cpp
```diff
@@ -0,0 +1,497 @@
...
+    }
+
+    item = (THRequestItem *)av_mallocz(sizeof(THRequestItem));
+    if (!item) {
+        goto fail;
+    }
+    item->lltask = NULL;
+    item->infer_request = th_create_inference_request();
+    if (!item->infer_request) {
+        av_log(NULL, AV_LOG_ERROR, "Failed to allocate memory for TensorFlow inference request\n");
```
wenbinc-Bin: This log should be "Torch", you forget to replace the backend's name.

libavfilter/dnn/dnn_io_proc.c
wenbinc-Bin: tmp_data is not freed.

libavfilter/dnn/dnn_io_proc.c
```diff
@@ -27,6 +27,12 @@
 int ff_proc_from_dnn_to_frame(AVFrame *frame, DNNData *output, void *log_ctx)
 {
     struct SwsContext *sws_ctx;
+    int frame_size = frame->height * frame->width;
+    int linesize[3];
+    linesize[0] = frame->linesize[0];
```
wenbinc-Bin, line 37: why just set the linesize[0]? yuv420p should have 3 planes.

libavfilter/dnn/dnn_io_proc.c
```diff
@@ -101,6 +147,13 @@ int ff_proc_from_frame_to_dnn(AVFrame *frame, DNNData *input, void *log_ctx)
 {
     struct SwsContext *sws_ctx;
     int bytewidth = av_image_get_linesize(frame->format, frame->width, 0);
+    int frame_size = frame->height * frame->width;
+    int linesize[3];
+    linesize[0] = frame->linesize[0];
```
wenbinc-Bin, line 161: Same as above. Why only set linesize[0].

libavfilter/dnn/dnn_io_proc.c
Semmer2: Actually, the SR model supported in openvino & tensorflow backend, only accept Y  channel, the U/V channel was scaled up by FFmpeg software scale.
Only libtorch SR model, which is the model we are working on, scale all three channels of RGB of image.
But you do alert me, I will check if the OpenVINO&Tensorflow will be influenced by this change.

libavfilter/dnn/dnn_io_proc.c
Semmer2: Sorry, I missed this part.

Semmer2:
@wenbinc-Bin Hi Wenbin, I have dropped out all the patches related to adding RGBP pixel format. I used the tricky way to do IO processing when do inference as we talk this afternoon. Thank you for review~😀

libavfilter/dnn/dnn_backend_torch.cpp
```diff
@@ -0,0 +1,569 @@
...
+    } else {
+        av_log(ctx, AV_LOG_ERROR, "Not supported device:\"%s\"\n", ctx->options.device_name);
+        return NULL;
+    }
+
+    try {
+        th_model->jit_model = torch::jit::load(model_filename, device);
+    } catch (const c10::Error& e) {
+        av_log(ctx, AV_LOG_ERROR, "Failed to load torch model\n");
+        return NULL;
```
wenbinc-Bin: Above three return should be "return fail;"?

libavfilter/dnn/dnn_backend_torch.cpp
```diff
@@ -0,0 +1,569 @@
...
+    ret = ff_queue_push_back(th_model->task_queue, task);
+    if (ret < 0) {
+        av_freep(&task);
+        av_log(ctx, AV_LOG_ERROR, "unable to push back task_queue.\n");
+        return ret;
+    }
+
+    ret = extract_lltask_from_task(task, th_model->lltask_queue);
+    if (ret != 0) {
+        av_freep(&task);
```
wenbinc-Bin: Is this av_freep(&task) needed?

libavfilter/dnn/dnn_backend_torch.cpp
```diff
@@ -0,0 +1,569 @@
...
+        av_log(ctx, AV_LOG_ERROR, "unable to extract last level task from task.\n");
+        return ret;
+    }
+
+    request = (THRequestItem *)ff_safe_queue_pop_front(th_model->request_queue);
+    if (!request) {
+        av_freep(&task);
```
wenbinc-Bin: same as above.

libavfilter/dnn/dnn_backend_torch.cpp
Semmer2: I think we should return NULL here if the load model failed. Since the function, which called this function, will check its' return value NULL or not, in dnn_filter_common.c:93.

libavfilter/dnn/dnn_backend_torch.cpp
Semmer2: It is needed, since we allocated the task memory manually in line481. So we need to free it manually as well? Hope I did not misunderstand you.

libavfilter/dnn/dnn_backend_torch.cpp
wenbinc-Bin: sorry, It is typo. I mean "goto fail;" because model and th_model need to be freed.

libavfilter/dnn/dnn_backend_torch.cpp
wenbinc-Bin: "task" is pushed into th_model->task_queue at line 494. If you return error, program will call release function ff_dnn_free_model_th() and it will be freed in this function. You can see tensorflow's implementation https://github.com/intel-media-ci/ffmpeg/blob/dd99d34d67e2612a8f133f8a86db9f64b4dfa20d/libavfilter/dnn/dnn_backend_tf.c#L1178

PR #497, opened 2022-01-20
libavcodec/qsvenc: fix a memory leak problem

wenbinc-Bin:
"qf->frame" ref to input frame but it isn't released. av_frame_unref()
is added before refering qf->frame to new frame to make sure the previous
reference is released.


xhaihao (approved):

wenbinc-Bin:
submit to: https://patchwork.ffmpeg.org/project/ffmpeg/patch/20220120071343.187817-1-wenbin.chen@intel.com/

PR #504, opened 2022-02-19
Error Specificity in DNN Module

shubhanshu02:
This patchset introduces error codes from `libavutil\error.h` in the Deep Neural Networks module to enhance the error handling.

### Usage

- `DNN_GENERIC_ERROR` for backend framework-specific errors
- `AVERROR(ENOSYS)` for missing features

### DIFF
--- a/libavfilter/dnn/dnn_backend_openvino.c
+++ b/libavfilter/dnn/dnn_backend_openvino.c
@@ -191,7 +191,7 @@ static int fill_model_input_ov(OVModel *ov_model, OVRequestItem *request)
     }
     ie_blob_free(&input_blob);
 
-    return DNN_SUCCESS;
+    return 0;
 }
 
 static void infer_completion_callback(void *args)
@@ -303,7 +303,7 @@ static void infer_completion_callback(void *args)
 
 static int init_model_ov(OVModel *ov_model, const char *input_name, const char *output_name)
 {
-    int ret = DNN_SUCCESS;
+    int ret = 0;
     OVContext *ctx = &ov_model->ctx;
     IEStatusCode status;
     ie_available_devices_t a_dev;
@@ -433,7 +433,7 @@ static int init_model_ov(OVModel *ov_model, const char *input_name, const char *
         goto err;
     }
 
-    return DNN_SUCCESS;
+    return 0;
 
 err:
     ff_dnn_free_model_ov(&ov_model->model);
@@ -444,7 +444,7 @@ static int execute_model_ov(OVRequestItem *request, Queue *inferenceq)
 {
     IEStatusCode status;
     LastLevelTaskItem *lltask;
-    int ret = DNN_SUCCESS;
+    int ret = 0;
     TaskItem *task;
     OVContext *ctx;
     OVModel *ov_model;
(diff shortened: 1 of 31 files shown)


libavfilter/dnn_filter_common.c
```diff
@@ -106,18 +106,18 @@ int ff_dnn_set_classify_post_proc(DnnContext *ctx, ClassifyPostProc post_proc)
     return 0;
 }
 
-DNNReturnType ff_dnn_get_input(DnnContext *ctx, DNNData *input)
+int ff_dnn_get_input(DnnContext *ctx, DNNData *input)
```
guoyejun, line 109: will there be build warning since ctx->model->get_input still returns DNNReturnType?

anyway, i think this is a good idea to decouple the change.

libavfilter/dnn_interface.h
```diff
@@ -30,6 +30,8 @@
 #include "libavutil/frame.h"
 #include "avfilter.h"
 
+#define DNN_GENERIC_ERROR FFERRTAG('D','N','N','!')
+
 typedef enum {DNN_SUCCESS, DNN_ERROR} DNNReturnType;
```
guoyejun: how about to remove DNNReturnType?

returns 0 for success, and negative values for errors.

shubhanshu02: Sure. That's seems better since `DNN_ERROR` isn't used anywhere after these changes.

libavfilter/dnn_filter_common.c
shubhanshu02: I didn't see any build warning related to `libavfilter` after this commit (no warning earlier also).

Decouple the change means?

guoyejun: i mean to separate the changes into small commits.

shubhanshu02: Oh, got it! Currently, I have migrated one module per commit which is a bigger change. Migrating one function each commit instead will lead to smaller commits.

guoyejun: it's ok for the current patch set

shubhanshu02: Oh, okay. :smile:

libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -495,22 +515,22 @@ static DNNReturnType get_input_ov(void *model, DNNData *input, const char *input
...
-        return DNN_ERROR;
+        return DNN_GENERIC_ERROR;
     }
 
     for (size_t i = 0; i < model_input_count; i++) {
         status = ie_network_get_input_name(ov_model->network, i, &model_input_name);
         if (status != OK) {
             av_log(ctx, AV_LOG_ERROR, "Failed to get No.%d input's name\n", (int)i);
-            return DNN_ERROR;
+            return DNN_GENERIC_ERROR;
```
Semmer2, line 525: The `status` was returned by `ie_network_get_intput_name`, but I think the error type `AVERROR(EINVAL)` would be better.

Semmer2:
LGTM, if @guoyejun has no more comments, its' OK to send them out. :)

shubhanshu02:
Thanks for reviewing. Can you also check the tests' logs that are failing - `ubuntu/focal` and `ubuntu/bionic`?

Semmer2:
> Thanks for reviewing. Can you also check the tests' logs that are failing - `ubuntu/focal` and `ubuntu/bionic`?

We believe it's a CI system issue, not from your code. It's OK to send them to maillist.

shubhanshu02:
Sure, I'll send the patchset to the mailing list. Thanks.

shubhanshu02:
Thanks for reviewing.

PR #508, opened 2022-02-23
Vulkan nv12

wenbinc-Bin:


wenbinc-Bin:
@xhaihao I try to fix nv12 and nv21 problem like this. Any suggestion? If no I will send to community.
Thanks
Wenbin

xhaihao:
How about change the commit log in the 2nd commit to: "The UV plane of NV21 should be mapped to GR88" ? We can't map GR88 to NV21.

wenbinc-Bin:
Thanks for review. I will update it.

libavutil/hwcontext_vulkan.c
```diff
@@ -3290,8 +3290,13 @@ static void vulkan_unmap_to_drm(AVHWFramesContext *hwfc, HWMapDescriptor *hwmap)
...
 }
 
-static inline uint32_t vulkan_fmt_to_drm(VkFormat vkfmt)
+static inline uint32_t vulkan_fmt_to_drm(VkFormat vkfmt,
+                                         enum AVPixelFormat sw_format)
 {
+    if (vkfmt == VK_FORMAT_R8G8_UNORM &&
+       (sw_format == AV_PIX_FMT_NV21 || sw_format == AV_PIX_FMT_NV42))
+            return DRM_FORMAT_GR88;
+
```
xhaihao, line 3303: How about using code below ? If so, we needn't swap the positions in the list

```
if (vkfmt == VK_FORMAT_R8G8_UNORM &&
       (sw_format == AV_PIX_FMT_NV12 || sw_format == AV_PIX_FMT_NV24))
            return DRM_FORMAT_RG88;
```

xhaihao: Is a similar check needed for `VK_FORMAT_R16G16_UNORM` ?

libavutil/hwcontext_vaapi.c
```diff
@@ -992,7 +992,7 @@ static const struct {
 } vaapi_drm_format_map[] = {
 #ifdef DRM_FORMAT_R8
     DRM_MAP(NV12, 2, DRM_FORMAT_R8,  DRM_FORMAT_RG88),
-    DRM_MAP(NV12, 2, DRM_FORMAT_R8,  DRM_FORMAT_GR88),
+    DRM_MAP(NV21, 2, DRM_FORMAT_R8,  DRM_FORMAT_GR88),
```
xhaihao: Mapping UV plane of NV12 to DRM_FORMAT_GR88 is WA, and I think we needn't add support for NV21 and  should remove `DRM_MAP(NV12, 2, DRM_FORMAT_R8,  DRM_FORMAT_GR88)` after applying the first commit.

BTW so far we don't have a driver to support NV21

libavutil/hwcontext_vulkan.c
wenbinc-Bin: VK_FORMAT_R16G16_UNORM will be mapped to DRM_FORMAT_GR1616 which is wrong so I swap it's position and I swap R8G8 as well.
I see there are only P010 and P016 using VK_FORMAT_R16G16_UNORM, and DRM_FORMAT_GR1616 is not used, so I think we don't need to check R16G16.

libavutil/hwcontext_vaapi.c
wenbinc-Bin: Ok, I will delete this line.

libavutil/hwcontext_vulkan.c
xhaihao: If so, how about to remove `{ DRM_FORMAT_GR1616,   VK_FORMAT_R16G16_UNORM   },` from this list ?

xhaihao:
typo in commit log:
s/prblem/problem

xhaihao:
You fixed both r8g8 and r16g16, could you update the commit log of the first commit ?

wenbinc-Bin:
The commit log is updated.

xhaihao (approved):

wenbinc-Bin:
submit to: https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=6329

PR #519, opened 2022-03-17
libavcodec/hevc_mp4toannexb_bsf: insert extradata before non-AUD unit

xhaihao:
It is possible that an IRAP frame in input AVPacket contains VPS, SPS
and PPS, and these headers should take effect. However the prepended
extradata might override these headers. This patch insert extradata
before non-AUD unit, hence VPS, SPS and PPS from the input AVPacket will
take effect if they are present.

This should fix #7799


xhaihao:
For testing only

xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=6220

PR #520, opened 2022-03-18
libavcodec/cbs_av1: Add size check before parse obu

wenbinc-Bin:
cbs_av1_write_unit() check pbc size after parsing obu frame, and return
AVERROR(ENOSPC) if pbc is small. pbc will be reallocated and this obu
frame will be parsed again, but this may cause error because
CodedBitstreamAV1Context has already been updated, for example
ref_order_hint is updated and will not match this obu frame. Now size
check is added before parsing obu frame to avoid this error.

### DIFF
--- a/libavcodec/cbs_av1.c
+++ b/libavcodec/cbs_av1.c
@@ -1075,6 +1075,9 @@ static int cbs_av1_write_obu(CodedBitstreamContext *ctx,
         put_bits32(pbc, 0);
     }
 
+    if (8 * unit->data_size > put_bits_left(pbc))
+        return AVERROR(ENOSPC);
+
     td = NULL;
     start_pos = put_bits_count(pbc);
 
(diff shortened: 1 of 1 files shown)


xhaihao:
May we remove https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/cbs_av1.c#L1199-L1200 after apply this patch ?

wenbinc-Bin:
Before parsing obu frames we don't know the exact header size. I think we should keep this code in case the real size requirement is larger than we expect.

wenbinc-Bin:
Can we add a log here?. Tell user that "the size is bigger than our expectation, reallocate may lead to parse error"

xhaihao (approved):
LGTM

wenbinc-Bin:
suhbmit to: https://patchwork.ffmpeg.org/project/ffmpeg/patch/20220323072507.584505-1-wenbin.chen@intel.com/

PR #523, opened 2022-03-23
qsvenc_{hevc,h264}: add scenario option

xhaihao:
Use may get better experience after providing corresponding senario hint
to encoder for the encoding session. See [1] for scenario support in the
SDK

[1]https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#scenarioinfo


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=7201

PR #541, opened 2022-05-26
qsv: check for libmfx.pc instead of mfx.pc

xhaihao:
This fixed the regression caused by commit 478e1a98a


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=6697

PR #546, opened 2022-06-02
avcodec/qsvenc: make QSV encoder encode VAAPI and D3D11 frames directly

tong1wu:
QSV encoder is able to encode VAAPI and D3D11 frames directly without
hwmap transformation. This patch adds support for qsv encoder to accept
VAAPI and D3D11 pixel formats as input.

Signed-off-by: Tong Wu <tong1.wu@intel.com>

### DIFF
--- a/libavcodec/qsvenc.c
+++ b/libavcodec/qsvenc.c
@@ -524,7 +524,9 @@ static int check_enc_param(AVCodecContext *avctx, QSVEncContext *q)
 
 static int init_video_param_jpeg(AVCodecContext *avctx, QSVEncContext *q)
 {
-    enum AVPixelFormat sw_format = avctx->pix_fmt == AV_PIX_FMT_QSV ?
+    enum AVPixelFormat sw_format = avctx->pix_fmt == AV_PIX_FMT_QSV ||
+                                   avctx->pix_fmt == AV_PIX_FMT_VAAPI ||
+                                   avctx->pix_fmt == AV_PIX_FMT_D3D11 ?
                                    avctx->sw_pix_fmt : avctx->pix_fmt;
     const AVPixFmtDescriptor *desc;
     int ret;
@@ -591,7 +593,9 @@ static int init_video_param_jpeg(AVCodecContext *avctx, QSVEncContext *q)
 
 static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
 {
-    enum AVPixelFormat sw_format = avctx->pix_fmt == AV_PIX_FMT_QSV ?
+    enum AVPixelFormat sw_format = avctx->pix_fmt == AV_PIX_FMT_QSV ||
+                                   avctx->pix_fmt == AV_PIX_FMT_VAAPI ||
+                                   avctx->pix_fmt == AV_PIX_FMT_D3D11 ?
                                    avctx->sw_pix_fmt : avctx->pix_fmt;
     const AVPixFmtDescriptor *desc;
     float quant;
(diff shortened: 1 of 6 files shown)


tong1wu:
@xhaihao @feiwan1 @wenbinc-Bin @xuguangxin Could you please review this patch thanks.

wenbinc-Bin:
How about to use `#define CONFIG_VAAPI` and `#define CONFIG_D3D11VA` to control the code. And there is a conflicts, maybe you need to update your code.

tong1wu:
> How about to use `#define CONFIG_VAAPI` and `#define CONFIG_D3D11VA` to control the code. And there is a conflicts, maybe you need to update your code.

Thanks, will update

tong1wu:
> How about to use `#define CONFIG_VAAPI` and `#define CONFIG_D3D11VA` to control the code. And there is a conflicts, maybe you need to update your code.

Have updated to solve the conflict. BTW do we really need to add macros to protect the pix_fmt enums as they have same behavior? If so we have to write the same codes twice in ff_qsv_enc_init.

wenbinc-Bin:
> > How about to use `#define CONFIG_VAAPI` and `#define CONFIG_D3D11VA` to control the code. And there is a conflicts, maybe you need to update your code.
> 
> Have updated to solve the conflict. BTW do we really need to add macros to protect the pix_fmt enums as they have same behavior? If so we have to write the same codes twice in ff_qsv_enc_init.

You are right. There are no vaapi or d3d specific data struct used, so the macro definition is not necessary.

xhaihao:
av_hwframe_map is actually called, so the commit log is wrong. 

```
QSV encoder is able to encode VAAPI and D3D11 frames directly without
hwmap transformation. 
```

libavcodec/qsvenc.c
```diff
@@ -1247,7 +1251,30 @@ int ff_qsv_enc_init(AVCodecContext *avctx, QSVEncContext *q)
...
+            AVBufferRef *derive_frames_ref = NULL;
+            ret = av_hwdevice_ctx_create_derived(&derive_device_ref,
+                                                 AV_HWDEVICE_TYPE_QSV, frames_ctx->device_ref, 0);
+            if (ret < 0) {
+                av_log(avctx, AV_LOG_ERROR, "Failed to derive QSV device context: %d.\n", ret);
+                return ret;
+            }
+            ret = av_hwframe_ctx_create_derived(&derive_frames_ref,
+                                                AV_PIX_FMT_QSV, derive_device_ref, avctx->hw_frames_ctx, 0);
+            if (ret < 0) {
```
xhaihao, line 1267: av_buffer_unref() should be call here to unref derive_device_ref.

libavcodec/qsvenc.c
```diff
@@ -1437,10 +1464,25 @@ static int submit_frame(QSVEncContext *q, const AVFrame *frame,
...
+                return ret;
+        } else {
+            qf->frame->format = AV_PIX_FMT_QSV;
+            qf->frame->hw_frames_ctx = av_buffer_ref(q->avctx->hw_frames_ctx);
+            if (!qf->frame->hw_frames_ctx) {
+                return AVERROR(ENOMEM);
+            }
+            ret = av_hwframe_map(qf->frame, frame, 0);
+            if (ret < 0) {
+                av_log(q->avctx, AV_LOG_ERROR, "Failed to map to QSV frames\n");
```
xhaihao, line 1480: call av_buffer_unref to unref qf->frame->hw_frames_ctx and return the error code.

tong1wu: Thanks, will update sooner

xhaihao (approved):
LGTM

xhaihao:
See https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=6763

PR #560, opened 2022-07-20
avcodec/av1_vaapi: Do not set invalid reference surface for each key …

feiwan1:
…frame

This may have risk to mess up reference list of backend driver.
Especially for a key frame that doesn't refresh all reference list.


feiwan1:
May relate to media-driver issue. Close this PR.

PR #564, opened 2022-08-03
Vp9

feiwan1:


libavcodec/decode.c
```diff
@@ -1202,6 +1229,15 @@ int ff_get_format(AVCodecContext *avctx, const enum AVPixelFormat *fmt)
         return AV_PIX_FMT_NONE;
 
     for (;;) {
+        if (avctx->internal->hwaccel_priv_data &&
+            avctx->hwaccel->caps_internal & HWACCEL_CAP_INTERNAL_ALLOC) {
+            err = hwaccel_realloc_surface(avctx);
```
xhaihao, line 1234: Is a flush needed before realllocating surface ?

feiwan1: > Is a flush needed before realllocating surface ?

The previous surface and hw_frames_ctx is stored/referenced by each AVFrame. So just need to create new hw_frames_ctx, and the new AVFrame will pick up surface from it. The old AVFrame still keep there with old surface if it is put into reference list. So don't need to flush it.

libavcodec/decode.c
```diff
@@ -1156,6 +1156,26 @@ static void hwaccel_uninit(AVCodecContext *avctx)
...
+        for (i = 0;; i++) {
+            hw_config = ffcodec(avctx->codec)->hw_configs[i];
+            if (!hw_config)
+                break;
+            if (hw_config->public.pix_fmt == fmt)
+                break;
+        }
+    } else {
+        hw_config = NULL;
+    }
```
xhaihao, line 1174: the else statement is not needed if hw_config is set to NULL when declaring it.

xhaihao: > the else statement is not needed if hw_config is set to NULL when declaring it.

Please ignore this comment.

xhaihao (approved):
LGTM

feiwan1:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=7206

PR #569, opened 2022-08-12
lavc/vaapi_encode_h265: correct sps syntax

feiwan1:
Get log2_min_luma_coding_block_size_minus3/log2_diff_max_min_luma_coding_block_size
from ctu and min cb size. And change default min_cb_size to 8 so that
keep log2_diff_max_min_luma_coding_block_size as its default value.

### DIFF
--- a/libavcodec/vaapi_encode_h265.c
+++ b/libavcodec/vaapi_encode_h265.c
@@ -430,12 +430,14 @@ static int vaapi_encode_h265_init_sequence_params(AVCodecContext *avctx)
             vps->vps_max_latency_increase_plus1[i];
     }
 
+    sps->log2_min_luma_coding_block_size_minus3 =
+        ff_ctz(priv->min_cb_size) - 3;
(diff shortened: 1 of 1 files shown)


libavcodec/vaapi_encode_h265.c
```diff
@@ -1209,14 +1207,14 @@ static av_cold int vaapi_encode_h265_get_encoder_caps(AVCodecContext *avctx)
 
     if (!priv->ctu_size) {
         priv->ctu_size     = 32;
-        priv->min_cb_size  = 16;
+        priv->min_cb_size  = 8;
```
xhaihao, line 1210: maybe this value doesn't work for the legacy driver.

feiwan1: This change doesn't affact any setting to libva. Maybe the new fail may caused by changing surface align from 16 to 32. I will check it.

xhaihao:
Please feel free to reopen it if this PR is still needed.

PR #573, opened 2022-08-18
qsvenc: support the combination of LA and CBR

guanjunz:
add LA CBR supports


libavcodec/qsvenc.c
```diff
@@ -756,6 +756,13 @@ static int init_video_param(AVCodecContext *avctx, QSVEncContext *q)
         q->param.mfx.TargetKbps  = target_bitrate_kbps / brc_param_multiplier;
         q->extco2.LookAheadDepth = q->look_ahead_depth;
         q->param.mfx.BRCParamMultiplier = brc_param_multiplier;
+        q->param.mfx.BufferSizeInKB = buffer_size_in_kilobytes / brc_param_multiplier;
+        q->param.mfx.InitialDelayInKB = initial_delay_in_kilobytes / brc_param_multiplier;
+        q->param.mfx.MaxKbps = max_bitrate_kbps / brc_param_multiplier;
+        if (avctx->bit_rate == avctx->rc_max_rate)
+        {
+            q->param.mfx.RateControlMethod = MFX_RATECONTROL_CBR;
```
xhaihao: Is there a way to fix this in select_rc_mode() ?

xhaihao:
You changed qsvenc.c only, so please use 'qsvenc:' instead of 'qsvenc_{hevc,h264}' in the log

libavcodec/qsvenc.c
guanjunz: I think what you suggest is more reasonable. There is a problem to handle 
if (q->extbrc) {
            q->extco2.LookAheadDepth = q->look_ahead_depth;
        }
and 
LA what we need

xhaihao:
@guanjunz Is this PR still needed ? If yes, could you please update your patch ?

PR #580, opened 2022-09-06
lavu/hwcontext_qsv: restrict the number of threads for the main session to 4

xhaihao:
If we don't restrict the number of threads, the SDK will create a thread
for each CPU. Too many threads are useless for hardware accelerated
decoding, encoding and processing because CPU does few things.


dvrogozh (commented):
Why you don't do the same for VP and qsv.c?
```
libavcodec/qsv.c:604:    sts = MFXInitEx(init_par, &session);
libavfilter/qsvvpp.c:991:    sts = MFXInit(implementation, pver, &session);
libavutil/hwcontext_qsv.c:1021:    sts = MFXInit(implementation, &ver, &session);
libavutil/hwcontext_qsv.c:1040:    sts = MFXInit(implementation, &ver, &session);
```

xhaihao:
`MFXInit` is not used any more for oneVPL path.

xhaihao:
Won't upstream.

PR #591, opened 2022-10-24
libavcodec/qsvenc_hevc: add tier option

nyanmisaka:
Without this change, MSDK always defaults the HEVC tier to MAIN if the `-level` is specified.

In addition, I set the default value of tier option to HIGH if the level >= 4.0 is specified, which has better tolerance for bitrate. This is because I observed a situation where the user does not know the max bitrate that the default MAIN tier can carry when setting the bitrate and level, which usually results in blocky videos due to the exceeded bitrate. For example, HEVC Main Level 5.0, Main Tier can only carry 25M while High Tier can carry 100M as per the HEVC specs.

I'd like to know your opinion on this.

### DIFF
--- a/doc/encoders.texi
+++ b/doc/encoders.texi
@@ -3654,6 +3654,15 @@ Set the encoding profile (scc requires libmfx >= 1.32).
 @item scc
 @end table
 
+@item @var{tier}
+Set the encoding tier (only level >= 4 can support high tier).
+This option only takes effect when the level option is specified.
+
+@table @samp
+@item main
+@item high
+@end table
+
 @item @var{gpb}
 1: GPB (generalized P/B frame)
 
--- a/libavcodec/qsvenc_hevc.c
+++ b/libavcodec/qsvenc_hevc.c
@@ -258,6 +258,9 @@ static const AVOption options[] = {
 #if QSV_VERSION_ATLEAST(1, 32)
     { "scc",     NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_PROFILE_HEVC_SCC     }, INT_MIN, INT_MAX,     VE, "profile" },
 #endif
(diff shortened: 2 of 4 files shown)


libavcodec/qsvenc_hevc.c
```diff
@@ -258,6 +258,9 @@ static const AVOption options[] = {
 #if QSV_VERSION_ATLEAST(1, 32)
     { "scc",     NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_PROFILE_HEVC_SCC     }, INT_MIN, INT_MAX,     VE, "profile" },
 #endif
+    { "tier",    NULL, OFFSET(qsv.tier), AV_OPT_TYPE_INT, { .i64 = MFX_TIER_HEVC_HIGH }, MFX_TIER_HEVC_MAIN, MFX_TIER_HEVC_HIGH, VE, "tier" },
```
xhaihao: You should add the help text.

doc/encoders.texi
```diff
@@ -3654,6 +3654,14 @@ Set the encoding profile (scc requires libmfx >= 1.32).
 @item scc
 @end table
 
+@item @var{tier}
+Set the encoding tier.
```
xhaihao: Could you add description in details ? According to your patch, It only takes effect for level >= 4.

xhaihao:
Could you add detailed explanation about your change in the commit log when sending your patches to FFmpeg ML ?

libavcodec/qsvenc_hevc.c
nyanmisaka: Done.

doc/encoders.texi
nyanmisaka: https://en.wikipedia.org/wiki/High_Efficiency_Video_Coding_tiers_and_levels

According to the HEVC specs, only level >= 4 can support HIGH tier.
I make a check here to avoid wrong settings.

xhaihao (approved):
LGTM, please send your patchset toFFmpeg mailing list, thx!

nyanmisaka:
Patches have been sent to the mailing list. This can be closed for now.
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=7848

PR #593, opened 2022-10-26
libavcodec/qsvenc: Add skip_frame support to qsvenc

wenbinc-Bin:
Add skip_frame support to qsvenc. Use per-frame metadata "qsv_skip_frame" to control it. skip_frame option defines the behavior of skip_frame.
no_skip: Frame skipping is disabled.
insert_dummy: Encoder inserts into bitstream frame where all macroblocks are encoded as skipped.
insert_nothing: Similar to insert_dummy, but encoder inserts nothing into bitstream.
brc_only: skip_frame metadata indicates the number of missed frames before the current frame.

### DIFF
--- a/doc/encoders.texi
+++ b/doc/encoders.texi
@@ -3564,6 +3564,24 @@ bitrate, @var{target_bitrate}, within the accuracy range @var{avbr_accuracy},
 after a @var{avbr_Convergence} period. This method does not follow HRD and the
 instant bitrate is not capped or padded.
 
+@item @var{skip_frame}
+Use per-frame metadata "qsv_skip_frame" to skip frame when encoding. This option
+defines the usage of this metadata.
+@table @samp
+@item no_skip
+Frame skipping is disabled.
+@item insert_dummy
+Encoder inserts into bitstream frame where all macroblocks are encoded as
+skipped.
+@item insert_nothing
+Similar to insert_dummy, but encoder inserts nothing into bitstream. The skipped
+frames are still used in brc. For example, gop still include skipped frames, and
+the frames after skipped frames will be larger in size.
+@item brc_only
+skip_frame metadata indicates the number of missed frames before the current
+frame.
+@end table
+
 @end table
 
 @subsection HEVC Options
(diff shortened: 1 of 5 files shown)


doc/encoders.texi
```diff
@@ -3564,6 +3564,22 @@ bitrate, @var{target_bitrate}, within the accuracy range @var{avbr_accuracy},
...
+Use per-frame metadata "qsv_skip_frame" to skip frame when encoding. This option
+defines the usage of this metadata.
+@table @samp
+@item no_skip
+Frame skipping is disabled.
+@item insert_dummy
+Encoder inserts into bitstream frame where all macroblocks are encoded as
+skipped.
+@item insert_nothing
+Similar to insert_dummy, but encoder inserts nothing into bitstream.
```
xhaihao: So the current frame is skipped in the stream, right ?

doc/encoders.texi
```diff
@@ -3564,6 +3564,22 @@ bitrate, @var{target_bitrate}, within the accuracy range @var{avbr_accuracy},
...
+@item brc_only
+skip_frame metadata indicates the number of missed frames before the current
```
xhaihao, line 3581: what is written into the stream ?

doc/encoders.texi
wenbinc-Bin: Yes.

doc/encoders.texi
wenbinc-Bin: According to my understanding, this affect bitrate for the rest frames. For example, if I set skip_frame for one frame, than the rest frames will be larger in size.

doc/encoders.texi
xhaihao: If so, why does user application not encode the current frame ?

doc/encoders.texi
xhaihao: What is the number of missed frames in this case ? My understanding is that skip_frame is a flag only.

wenbinc-Bin: Skipframe in co2 is flag which indicate the usage of SkipFrame in enc_ctrl..
Skipframe in enc_ctrl is flag when using insert_dummy and insert_nothing mode, but in brc_only mode it is the number of missed frame.

doc/encoders.texi
wenbinc-Bin: The skipped frames will still be used for brc. I added more explanation into doc.

doc/encoders.texi
xhaihao: Got it, thanks for clarifying it.

xhaihao (approved):
LGTM.

xhaihao:
see https://patchwork.ffmpeg.org/project/ffmpeg/patch/20221102081151.803641-1-wenbin.chen@intel.com/

PR #602, opened 2022-11-21
RFC: Add HEVC Rext decode support for d3d11va/dxva2 on Intel

nyanmisaka:
Since the HEVC Rext has been commonly supported on Intel IceLake and newer platforms and 4:2:2 has been used in many advanced video recorders, as well as the recently released [Chromium](https://github.com/chromium/chromium/commit/a790b3df6caa9015eb6554674635a53fd117b38a) has supported Rext on Intel too, I think it is worth to add this in d3d11va/dxva2 to accelerate their playback in third-party players with the help of [LavFilters](https://github.com/Nevcairiel/LAVFilters).

The essential patch 3/5 is taken from the [VLC player](https://code.videolan.org/videolan/vlc/-/blob/master/contrib/src/ffmpeg/0001-avcodec-dxva2_hevc-add-support-for-parsing-HEVC-Rang.patch) and it works fine for years but it seems that the [ffmpeg dev preferred](https://patchwork.ffmpeg.org/project/ffmpeg/patch/20200313102354.2500-1-robux4@ycbcr.xyz/) to have Microsoft to release a new header for HEVC Rext.

I have verified this patch locally, including its interop with QSV VPP.

I wonder if you guys can push this series into ffmpeg or work with Microsoft to release a new header for HEVC Rext?


tong1wu:
Currently HEVC Rext related GUIDs need support from Microsoft SDK headers. It looks like it's a better idea for us to connect with Microsoft for a new released header with those GUIDs defined. We will push this process, thanks.

PR #617, opened 2023-01-03
qsv: add preset target bitrate

mamingy2:
Add preset target bitrate for encoding instead of fixed bitrates.
This result is calculated by linear interpolation of data such as length, width and framerate.
The interpolation method is from MediaSDK sample_utils CalculateDefaultBitrate function.


mamingy2:
Hi @xhaihao @feiwan1 , could you help to review? It looks like I don't have the right to ask you for reviews in github PR page. Thanks!

PR #619, opened 2023-01-04
hwcontext_vaapi: Add support for vaGetDisplayWin32 initialization path

sivileri:
Starting from [libva v2.17](https://github.com/intel/libva/releases/tag/2.17.0), there's a new display node called `libva-win32`, allowing support for libva on Windows OS. A Windows VAAPI driver based on Direct3D 12 Video APIs is also available as part of [Mesa 22.3.0 release](https://docs.mesa3d.org/relnotes/22.3.0.html) or later, via their VA frontend and gallium d3d12 driver.

The changes in this PR add support for `hwcontext_vaapi` VAAPI initialization using `vaGetDisplayWin32`, also allowing for device (adapter) selection via "-device" command line argument. With this change, it's possible to run VAAPI workloads on FFMpeg on Windows. For example, using the Windows driver mentioned above, H264/HEVC/VP9/AV1 8/10bit Decode and H264/HEVC 8/10bit encode as well as some video processing effects are supported on ffmpeg/vaapi.

## Basic test instructions

1. Download [libva and mesa vaon12 VA driver]( https://www.nuget.org/packages/Microsoft.Direct3D.VideoAccelerationCompatibilityPack)
2. Rename .nupkg to .zip and uncompress the file.
3. From build\native\<arch>\bin\>, copy vaon12_drv_video.dll, dxil.dll, va.dll, va_win32.dll to c:\bin (or any preferred directory)
4. Add c:\bin to your PATH environment user variable
5. Set the following user environment variables
```
    // Prefix of VA driver dll name
    LIBVA_DRIVER_NAME=vaon12
    // Path to directory containing vaon12_drv_video.dll
    LIBVA_DRIVERS_PATH=C:\bin
```

6. Build ffmpeg from this PR source code with --enable-vaapi and copy the produced binaries to c:\bin
        a. You can find the associated dependencies lib/ and include/ folders on the same package described in (1).
8. Run ffmpeg/vaapi commands as usual like for example from a powershell terminal: 

### VAAPI H264 Decode + Encode
ffmpeg.exe -hwaccel vaapi -hwaccel_output_format vaapi -hwaccel_device 0 -i input_720p.h264 -vcodec h264_vaapi output.h264 -loglevel trace

### VAAPI H264 Decode + Encode with HW accelerated scale VPP effect
ffmpeg.exe -hwaccel vaapi -hwaccel_output_format vaapi -hwaccel_device 0 -i input_720p.h264 -vf 'scale_vaapi=w=1280:h=720' -vcodec h264_vaapi output.h264 -loglevel trace


sivileri:
I don't seem to have permissions to add reviewers, so tagging them here. Could a mantainer please add them to reviewers? Thanks!

@xhaihao @feiwan1 @xuguangxin @galinart

sivileri:
The configure argument "--enable-vaapi" should be added to the intel/ci/windows to add coverage for the new code here, but I don't seem to have access to the pipelines to add that, how could I help adding this argument flag?

libavutil/hwcontext_dxcore.cpp
```diff
@@ -0,0 +1,119 @@
+/*
```
dvrogozh: ffmpeg is written in C, not C++. They do prefer to work in a C style with DX API. For example:
https://github.com/FFmpeg/FFmpeg/blob/4a80db5fc2b1b134550ffbcb8fd3b7ce2ad734b3/libavcodec/dxva2.c#L180
So, a call looks as:
```
hr = IDirectXVideoDecoderService_GetDecoderRenderTargets((IDirectXVideoDecoderService *)decoder_service, &guid, &target_count, &target_list);
```

dvrogozh (commented):
I would add some description to commit message including short note how to get working driver and few cmdline examples.

libavutil/hwcontext_dxcore.cpp
sivileri: Unfortunately, DXCore does not have a C API available similar to the ones used in DXVA/D3D11VA. I separated the cpp code in this new file to avoid tainting the existing VAAPI C code. I see other .cpp files in the project, would this be an issue given that there's no pure C API available?

dvrogozh: I don't know. @galinart will probably be the best to comment since he worked with ffmpeg on Windows.

tong1wu:
If enable-vaapi and enable-libvpl are both specified, it will have compile error with this PR. https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsv_internal.h#L38. Since there's no va/va_drm.h on Windows and CONFIG_VAAPI now is true, compile errors are introduced.

Should be very careful with not only vaapi itself but also vaapi related parts such as libvpl vulkan opencl.

libavutil/hwcontext_dxcore.cpp
sivileri: @galinart kindly pinging on this :)

sivileri:
> If enable-vaapi and enable-libvpl are both specified, it will have compile error with this PR. https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsv_internal.h#L38. Since there's no va/va_drm.h on Windows and CONFIG_VAAPI now is true, compile errors are introduced.
> 
> Should be very careful with not only vaapi itself but also vaapi related parts such as libvpl vulkan opencl.

Thanks for pointing this out and for your patience, just coming back to the office. I'd be happy to include extra commits to fix these issues in this PR.

Looks like qsv_internal.h:38 should not need to include va_drm.h as vaGetDisplayDRM (the only defined thing in that header) is only used in ffmpeg\libavutil\hwcontext_vaapi.c which includes va_drm.h guarded by `#if HAVE_VAAPI_DRM`? Then creating a VAAPI device context should be transparent to the actual VAAPI backend (ie. X11, DRM, Win32, etc) by calling `av_hwdevice_ctx_create(..., AV_HWDEVICE_TYPE_VAAPI, ...)`?

For the other places that use CONFIG_VAAPI, they look fine to me as they refer to texture formats and generic VAAPI concepts like VASurfaceIDs and VADisplayIDs common across all VAAPI backends (X11, DRM, Win32, etc). There are some parts in vulkan that use CONFIG_VAAPI with DRM, but they do inside a CONFIG_LIBDRM ifdef.

Does this make sense to you?

sivileri:
> I would add some description to commit message including short note how to get working driver and few cmdline examples.

I've updated the PR description also with some basic testing steps. Please let me know if you have any issues or questions.

sivileri:
I don't seem to have permissions to re-request for all reviewers after new changes, could a mantainer please add them to reviewers? Thanks!

@xhaihao @feiwan1 @xuguangxin @galinart @tong1wu

tong1wu:
> > If enable-vaapi and enable-libvpl are both specified, it will have compile error with this PR. https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsv_internal.h#L38. Since there's no va/va_drm.h on Windows and CONFIG_VAAPI now is true, compile errors are introduced.
> > Should be very careful with not only vaapi itself but also vaapi related parts such as libvpl vulkan opencl.
> 
> Thanks for pointing this out and for your patience, just coming back to the office. I'd be happy to include extra commits to fix these issues in this PR.
> 
> Looks like qsv_internal.h:38 should not need to include va_drm.h as vaGetDisplayDRM (the only defined thing in that header) is only used in ffmpeg\libavutil\hwcontext_vaapi.c which includes va_drm.h guarded by `#if HAVE_VAAPI_DRM`? Then creating a VAAPI device context should be transparent to the actual VAAPI backend (ie. X11, DRM, Win32, etc) by calling `av_hwdevice_ctx_create(..., AV_HWDEVICE_TYPE_VAAPI, ...)`?
> 
> For the other places that use CONFIG_VAAPI, they look fine to me as they refer to texture formats and generic VAAPI concepts like VASurfaceIDs and VADisplayIDs common across all VAAPI backends (X11, DRM, Win32, etc). There are some parts in vulkan that use CONFIG_VAAPI with DRM, but they do inside a CONFIG_LIBDRM ifdef.
> 
> Does this make sense to you?



> > If enable-vaapi and enable-libvpl are both specified, it will have compile error with this PR. https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsv_internal.h#L38. Since there's no va/va_drm.h on Windows and CONFIG_VAAPI now is true, compile errors are introduced.
> > Should be very careful with not only vaapi itself but also vaapi related parts such as libvpl vulkan opencl.
> 
> Thanks for pointing this out and for your patience, just coming back to the office. I'd be happy to include extra commits to fix these issues in this PR.
> 
> Looks like qsv_internal.h:38 should not need to include va_drm.h as vaGetDisplayDRM (the only defined thing in that header) is only used in ffmpeg\libavutil\hwcontext_vaapi.c which includes va_drm.h guarded by `#if HAVE_VAAPI_DRM`? Then creating a VAAPI device context should be transparent to the actual VAAPI backend (ie. X11, DRM, Win32, etc) by calling `av_hwdevice_ctx_create(..., AV_HWDEVICE_TYPE_VAAPI, ...)`?
> 
> For the other places that use CONFIG_VAAPI, they look fine to me as they refer to texture formats and generic VAAPI concepts like VASurfaceIDs and VADisplayIDs common across all VAAPI backends (X11, DRM, Win32, etc). There are some parts in vulkan that use CONFIG_VAAPI with DRM, but they do inside a CONFIG_LIBDRM ifdef.
> 
> Does this make sense to you?

Yes it looks like we don't have to include va_drm.h for qsv_internal.h.

My other concerns:
1. Previously we regarded VAAPI as Linux only and this assumption appears in a lot of places. For example, https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsv_internal.h#L27 might need a new name. https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/hwcontext_qsv.c#L2123. This will make AV_HWDEVICE_TYPE_VAAPI as qsv's child_device_type when you don't explicitly specify child_device_type. Previously the default should be AV_HWDEVICE_TYPE_D3D11VA.

2. Besides qsv, opencl and vulkan have vaapi related code inside(CONFIG_VAAPI for vulkan and HAVE_OPENCL_VAAPI_INTEL for opencl). Could you please enable vaapi with them to see if there's other compile or logic problems?

I think after that we still need to run some tests in our CI to make sure there's no unexpected behaviors during execution.

sivileri:
@tong1wu 

> Yes it looks like we don't have to include va_drm.h for qsv_internal.h.

I've added a new commit for removing this include.

> My other concerns:
> 
> 1. Previously we regarded VAAPI as Linux only and this assumption appears in a lot of places. For example, https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/qsv_internal.h#L27 might need a new name. 

In another new commit, I renamed it to `AVCODEC_QSV_VAAPI_SESSION_HANDLE`. Please note that `ff_qsv_set_display_handle` sets some driver default names and AFAIK those are not available on Windows yet. The other places look well to me, as they just refer to a generic VAAPI device/session and are not assuming the backend (ie. win32, x11, drm, etc).  Is the (QSV on top of VAAPI using the vaon12 driver) scenario included in the CI tests you mentioned below?

> https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/hwcontext_qsv.c#L2123. This will make AV_HWDEVICE_TYPE_VAAPI as qsv's child_device_type when you don't explicitly specify child_device_type. Previously the default should be AV_HWDEVICE_TYPE_D3D11VA.

Great catch. I've swapped the if/else cases so it stays VAAPI as default for linux, and stays DXVA2/D3D11_VA default in Windows too even when VAAPI is present.

> 2. Besides qsv, opencl and vulkan have vaapi related code inside(CONFIG_VAAPI for vulkan and HAVE_OPENCL_VAAPI_INTEL for opencl). Could you please enable vaapi with them to see if there's other compile or logic problems?

I looked at that code and, for vulkan it seems ok as DRM assumptions for map/unmap are guarded by `#ifdef LIBDRM`. For opencl, looks good also, as all the logic is delegated onto the [cl_intel_va_api_media_sharing extension](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_va_api_media_sharing.txt) without assuming backend (ie. win32, x11, drm, etc). 

I successfully built vaapi, opencl, libvpl and vulkan together. For opencl, this fix is needed https://github.com/intel/libva/pull/680 as I was missing an exported function `vaDisplayIsValid`.

> I think after that we still need to run some tests in our CI to make sure there's no unexpected behaviors during execution.

Sounds good, are the qsv/opencl/vulkan scenarios with VAAPI included in this CI tests?

tong1wu:
> Sounds good, are the qsv/opencl/vulkan scenarios with VAAPI included in this CI tests?

On Windows currently our CI only covers ffmpeg-libvpl(d3d11 and dxva2 as child_device_type). Opencl and vulkan are not included. I think ffmpeg-vaapi(d3d12 backend) and ffmpeg-libvpl(vaapi as child_device_type) are needed in the future.

sivileri:
> > Sounds good, are the qsv/opencl/vulkan scenarios with VAAPI included in this CI tests?
> 
> On Windows currently our CI only covers ffmpeg-libvpl(d3d11 and dxva2 as child_device_type). Opencl and vulkan are not included. I think ffmpeg-vaapi(d3d12 backend) and ffmpeg-libvpl(vaapi as child_device_type) are needed in the future.

Gotcha @tong1wu . I've updated the instructions in the PR description for running ffmpeg-vaapi (layered on d3d12), please let me know if it works for you, should be feasible to add that to the CI to test that scenario? For ffmpeg-libvpl I don't have experience on how that works yet, but I'd be happy to help with trying it out, is there any documentation/command line samples I can follow? 

ps. I can't access the CI failures, I see intel/ci/ubuntu/jammy/pr/check and intel/ci/focal/pr/check failing, could you tell me what's the failing issue? Thanks

tong1wu:
> > > Sounds good, are the qsv/opencl/vulkan scenarios with VAAPI included in this CI tests?
> > 
> > 
> > On Windows currently our CI only covers ffmpeg-libvpl(d3d11 and dxva2 as child_device_type). Opencl and vulkan are not included. I think ffmpeg-vaapi(d3d12 backend) and ffmpeg-libvpl(vaapi as child_device_type) are needed in the future.
> 
> Gotcha @tong1wu . I've updated the instructions in the PR description for running ffmpeg-vaapi (layered on d3d12), please let me know if it works for you, should be feasible to add that to the CI to test that scenario? For ffmpeg-libvpl I don't have experience on how that works yet, but I'd be happy to help with trying it out, is there any documentation/command line samples I can follow?
> 
> ps. I can't access the CI failures, I see intel/ci/ubuntu/jammy/pr/check and intel/ci/focal/pr/check failing, could you tell me what's the failing issue? Thanks

The CI failures were caused by a FFmpeg upstream issue and should have nothing to do with this PR.

Actually ffmpeg-libvpl doesn't support vaapi on Windows for now(only d3d11va and dxva2). As a result, more changes should be made to raise errors when users manually set child_device_type to vaapi. 

Now the pre-check only has build check but doesn't have bat check which runs a bunch of cases and succeeds only all the cases are passed. We are pushing forward this process and it might still need some time. In fact I think you can try to upstream your patch set after we confirm our current components are not affected (ffmpeg-d3d11 and ffmpeg-libvpl:d3d11).

libavutil/hwcontext_dxcore.cpp
sivileri: @galinart @dvrogozh . To avoid any issues with C++ code given [the C code requirement in the ffmpeg developer guide](https://ffmpeg.org/developer.html#toc-C-language-features) I re-wrote the adapter selection logic in pure C following the [D3D11VA similar logic](https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/hwcontext_d3d11va.c#L581), please let me know if looks good to you. CC @tong1wu

sivileri:
> Actually ffmpeg-libvpl doesn't support vaapi on Windows for now(only d3d11va and dxva2). As a result, more changes should be made to raise errors when users manually set child_device_type to vaapi.

@tong1wu Makes sense. Based on [ffmpeg developer guidance](https://ffmpeg.org/developer.html#Submitting-patches-1), to try and make the patch as minimal and self-contained as possible for enabling `Windows ffmpeg-vaapi` functionality, I think it may be better that I just send to the ffmpeg community a first patch only with these two commits from this PR:
* [hwcontext_vaapi: Add support for vaGetDisplayWin32 initialization path](https://github.com/intel-media-ci/ffmpeg/pull/619/commits/578e117ce35c4c47871c5dca9f4ef9db2a4984a4)
* [qsv_internal.h: Remove unnecessary include va_drm.h](https://github.com/intel-media-ci/ffmpeg/pull/619/commits/78bf369cdf3d1a83722b0af766e0e0ec14fd27a5)
    *  Include this seemingly unrelated one to avoid build regressions when `enable-vaapi` and `enable-libvpl` are set with the first commit changes.

For other `ffmpeg-libvpl` related commits, I can create a separate PR with the other `qsv` commits, along with new ones to raise errors when users manually set child_device_type to vaapi. This would be a separate patch to send to the ffmpeg community after the first one for enabling `ffmpeg-vaapi` functionality only.

Any concerns with this?

> Now the pre-check only has build check but doesn't have bat check which runs a bunch of cases and succeeds only all the cases are passed. We are pushing forward this process and it might still need some time. In fact I think you can try to upstream your patch set after we confirm our current components are not affected (ffmpeg-d3d11 and ffmpeg-libvpl:d3d11).

Sounds great, please let me know once that's been confirmed. Will the [ffmpeg FATE tests](https://ffmpeg.org/fate.html) be run at some point during the validation of `ffmpeg-d3d11` and `ffmpeg-libvpl:d3d11`?

libavutil/hwcontext_dxcore.cpp
galinart: It is better to avoid cpp files but this is FFmpeg community decides if it is acceptable.

configure
```diff
@@ -6938,6 +6939,21 @@ test_cpp <<EOF && enable uwp && d3d11va_extralibs="-ldxgi -ld3d11"
...
 
+# vaapi_win32 requires linking directly to dxgi if not building for
+# the desktop api partition
+test_cpp <<EOF && enable uwp && vaapi_win32_extralibs="-ldxgi"
+#ifdef WINAPI_FAMILY
+#include <winapifamily.h>
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+#error desktop, not uwp
+#else
+// WINAPI_FAMILY_APP, WINAPI_FAMILY_PHONE_APP => UWP
```
dvrogozh, line 6951: is that commented out on purpose? why not a C-style comments?

libavutil/hwcontext_vaapi.c
```diff
@@ -1788,6 +1799,60 @@ static int vaapi_device_create(AVHWDeviceContext *ctx, const char *device,
...
+    if (!display && try_win32) {
+        // Try to create a display from the specified device, if any.
+        if (!device) {
+            display = vaGetDisplayWin32(NULL);
+        } else {
+            IDXGIFactory2 *pDXGIFactory = NULL;
+            IDXGIAdapter *pAdapter = NULL;
+            typedef HRESULT(WINAPI *PFN_CREATE_DXGI_FACTORY)(REFIID riid, void **ppFactory);
+#if !HAVE_UWP
+            HANDLE dxgilib = dlopen("dxgi.dll", 0);
```
dvrogozh: sorry for ignorance, but why dlopen here and FreeLibrary down below?

sivileri:
Ran successfully the FATE test `fate-hwdevice` to validate the changes in `hwdevice_vaapi.c`.

```
$ CC=gcc LD=ld make fate-hwdevice SAMPLES=fate-suite/
CC      libavutil/version.o
version.c
AR      libavutil/libavutil.a
LD      libavutil/tests/hwdevice.exe
TEST    hwdevice
```

libavutil/hwcontext_vaapi.c
```diff
@@ -1788,6 +1800,65 @@ static int vaapi_device_create(AVHWDeviceContext *ctx, const char *device,
...
+                                              (void **)&pDXGIFactory))) {
+                int adapter = atoi(device);
+                if (SUCCEEDED(IDXGIFactory2_EnumAdapters(pDXGIFactory,
+                                                         adapter,
+                                                         &pAdapter))) {
+                    DXGI_ADAPTER_DESC desc;
+                    if (SUCCEEDED(IDXGIAdapter2_GetDesc(pAdapter, &desc))) {
+                        av_log(ctx, AV_LOG_INFO,
+                              "Using device %04x:%04x (%ls).\n",
+                              desc.VendorId, desc.DeviceId, desc.Description);
```
dvrogozh: vendor, device and description are not unique identifiers of the adapter in case of the system with multiple adapters (of the same type). I suggest that we need to also print LUID.

libavutil/hwcontext_vaapi.c
```diff
@@ -1788,6 +1800,65 @@ static int vaapi_device_create(AVHWDeviceContext *ctx, const char *device,
...
+#if HAVE_VAAPI_WIN32
+    if (!display && try_win32) {
+        // Try to create a display from the specified device, if any.
+        if (!device) {
+            display = vaGetDisplayWin32(NULL);
+        } else {
+            IDXGIFactory2 *pDXGIFactory = NULL;
+            IDXGIAdapter *pAdapter = NULL;
+#if !HAVE_UWP
+            HANDLE dxgi = dlopen("dxgi.dll", 0);
```
dvrogozh, line 1812: why dlopen here and FreeLibrary down below?

libavutil/hwcontext_vaapi.c
sivileri: Great catch, I'll use `dlclose` to be consistent.

libavutil/hwcontext_vaapi.c
sivileri: See other comment related to this one, will use `dlclose` for consistency.

libavutil/hwcontext_vaapi.c
dvrogozh: Just curious: ffmpeg does provide dlopen/dlclose wrappers on Windows?

sivileri: Yes, in `"compat/w32dlfcn.h"` as ` hwcontext_d3d11va` uses.

configure
sivileri: Please see above at `line 6927`, a very similar check is done for the `d3d11va` backend. I think it should stay like this for consistency with those other existing checks.

dvrogozh: Ok. II am fine with that.

dvrogozh (dismissed):
from my perspective this looks good

tong1wu:
> For other `ffmpeg-libvpl` related commits, I can create a separate PR with the other `qsv` commits, along with new ones to raise errors when users manually set child_device_type to vaapi. This would be a separate patch to send to the ffmpeg community after the first one for enabling `ffmpeg-vaapi` functionality only.
> 
> Any concerns with this?
> 

I am afraid that it's not what you expected. In fact as long as you have vaapi dependency in your environment, vaapi will be enabled even if you don't specify --enable-vappi. As a result, users will see issues when they only set --enable-libvpl but have vaapi dependency in their environments.

tong1wu:
I tested ffmpeg-libvpl:d3d11 and ffmpeg-d3d11 with this PR on my local CI environment and did not see regressions.

xhaihao:
Please bump a new MINOR version in libavutil/version.h  and document your change in Changelog for this new feature.

sivileri:
> > For other `ffmpeg-libvpl` related commits, I can create a separate PR with the other `qsv` commits, along with new ones to raise errors when users manually set child_device_type to vaapi. This would be a separate patch to send to the ffmpeg community after the first one for enabling `ffmpeg-vaapi` functionality only.
> > Any concerns with this?
> 
> I am afraid that it's not what you expected. In fact as long as you have vaapi dependency in your environment, vaapi will be enabled even if you don't specify --enable-vappi. As a result, users will see issues when they only set --enable-libvpl but have vaapi dependency in their environments.

@tong1wu  Gotcha, thanks for clarifying. Then looks like I should send a patch to `ffmpeg-devel` with all the squashed changes on this PR, right?

> I tested ffmpeg-libvpl:d3d11 and ffmpeg-d3d11 with this PR on my local CI environment and did not see regressions.

Perfect @tong1wu . Before I go ahead and post this patch in the ffmpeg community, do you still feel like I should add in this PR new logic to raise errors when users manually set child_device_type to vaapi in libvpl? Or the testing you did validated this scenario with the code changes as-is?

xhaihao:
> @tong1wu Gotcha, thanks for clarifying. Then looks like I should send a patch to `ffmpeg-devel` with all the squashed changes on this PR, right?

You should squash some commits into one at least. E.g. the first 2 commits should be squashed into one commit only, otherwise the first commit will introduce a compiling error. We should make sure each commit is compilable after applying. 

Moreover we should make sure no error / regression is introduced after a commit is applied, so it would be better to squash the first 2 commits and the 4th commit into one. 

> 
> > I tested ffmpeg-libvpl:d3d11 and ffmpeg-d3d11 with this PR on my local CI environment and did not see regressions.
> 
> Perfect @tong1wu . Before I go ahead and post this patch in the ffmpeg community, do you still feel like I should add in this PR new logic to raise errors when users manually set child_device_type to vaapi in libvpl?

I think it is better to do so.

libavcodec/qsv_internal.h
```diff
@@ -24,10 +24,10 @@
 #include "config.h"
 
 #if CONFIG_VAAPI
-#define AVCODEC_QSV_LINUX_SESSION_HANDLE
+#define AVCODEC_QSV_VAAPI_SESSION_HANDLE
```
xhaihao: You  should make sure AVCODEC_QSV_LINUX_SESSION_HANDLE (or AVCODEC_QSV_VAAPI_SESSION_HANDLE) is not defined on Windows, otherwise the code below is wrong. 

```
#ifdef AVCODEC_QSV_LINUX_SESSION_HANDLE                                                                                                                                                        
    ret = ff_qsv_set_display_handle(avctx, qs);
    if (ret < 0)
        return ret;
#endif
```
It will try to set display handle even if user specifies d3d11va or dxva2 type.

sivileri: Will change in next push

sivileri:
@xhaihao Thanks for your comments. In the last pushed changes:
- I squashed most commits to have a single commit that doesn't break components. Edited the commit description with the combined changes.
- Only define `AVCODEC_QSV_LINUX_SESSION_HANDLE` for `CONFIG_VAAPI and !_WIN32`, so this won't be enabled for Windows builds even when libva-win32 is present, and the session handle **will not** be assigned.
- In `qsv_device_create`
    - Change the priority order to have VAAPI after DXVA2/D3D11VA when autoselecting the `child_device_type`
        - On linux shouldn't matter as DX based is not available. On Windows keeps the priority order as before this change.
    - Validate the `child_device_type` is not VAAPI on Windows and issue an error otherwise.

These changes passed `make fate-hwdevice`. Please let me know if the QSV tests still pass with the latest updated code.

CC @dvrogozh @tong1wu

tong1wu:
Tested again and did not find regressions for ffmpeg-d3d11 and ffmpeg-libvpl. LGTM

sivileri:
Thanks! Will close this MR and send patch to community as per README at https://github.com/intel/cartwheel-ffmpeg

sivileri:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230216131213.2326-1-sivileri@microsoft.com/

PR #625, opened 2023-02-09
libavcodec/qsvenc: Flush cached frames before reset encoder

wenbinc-Bin:
According to https://github.com/Intel-Media-SDK/MediaSDK/blob/master/doc/mediasdk-man.md#configuration-change. Before calling MFXVideoENCODE_Reset, The application needs to retrieve any cached frames in the SDK encoder.
A loop is added before MFXVideoENCODE_Reset to retrieve cached frames and add them to async_fifo, so that dynamic configuration works when async_depth > 1.


libavcodec/qsvenc.c
```diff
@@ -2466,6 +2414,66 @@ static int encode_frame(AVCodecContext *avctx, QSVEncContext *q,
...
+                continue;
+            q->param.ExtParam[q->param.NumExtParam++] = q->extparam_internal[i];
+        }
+    } else {
+        q->param.ExtParam    = q->extparam_internal;
+        q->param.NumExtParam = q->nb_extparam_internal;
+    }
+
+    //Flush codec before reset configuration.
+    while (ret != AVERROR(EAGAIN)) {
```
xhaihao, line 2463: Is this condition right when encode_frame() below returns AVERROR(EAGAIN) ?

wenbinc-Bin: According to documentation. SDK decide whether to start a new sequence and reset codec status.
I test it on my side: if I change bitrate, gop struct will not be influenced. If I change gop size, gop struct is changed and a new gop is started.
If I enable b_frames, mini gop struct will be change. A p_frame is inserted where the parameter is changed.

xhaihao (approved):

xhaihao (approved):
LGTM

wenbinc-Bin:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230213053931.534552-1-wenbin.chen@intel.com/

ZaquL:
About your https://github.com/FFmpeg/FFmpeg/commit/e530d38bbf5f58e3e22f87dff635a50e267abb77

But RGB is the same as YCbCr 444. You should just tag it as matrix identity, full range, sRGB transfer, BT.709 primaries after encoding 444 frame that you pass as YCbCr 444 frame.

wenbinc-Bin:
You can do this when encoding system memory but you cannot do this when encoding video memory.
I think qsv should better align with VPL.

ValZapod:
>but you cannot do this when encoding video memory.

As in when you transcode RGB video?

ZaquL:
As in when you transcode RGB video?

wenbinc-Bin:
Sorry, I don't quite get your question. I try to answer it.
When you map frame from 3D framework or you want to build a full GPU transcode pipeline, the frames need to be uploaded to GPU memory before they are sent to encoder. The qsvenc will receive MfxSurface which points to vaSurface or D3dSurface. You need to change its format descriptor to encode them as YUV, which is not encouraged.

PR #627, opened 2023-02-24
lavc/vp9: Add RGB* formats for VAAPI hwaccel

feiwan1:


libavcodec/vp9.c
```diff
@@ -239,6 +239,13 @@ static int update_size(AVCodecContext *avctx, int w, int h)
         case AV_PIX_FMT_YUV444P12:
 #if CONFIG_VP9_VAAPI_HWACCEL
             *fmtp++ = AV_PIX_FMT_VAAPI;
+#endif
+            break;
+        case AV_PIX_FMT_GBRP:
+        case AV_PIX_FMT_GBRP10:
+        case AV_PIX_FMT_GBRP12:
+#if CONFIG_VP9_VAAPI_HWACCEL
+            *fmtp++ = AV_PIX_FMT_VAAPI;
```
xhaihao, line 248: GBRP and varieties are planar formats, which driver supports them ? I can't find these formats in https://github.com/intel/media-driver/blob/master/docs/media_features.md#supported-encoding-input-format-and-max-resolution.

feiwan1: The RGB* is parserd from frame header according to color_space flag. Even driver may not support RGB* format directly, it may choose the best matched format base on queried surface formats result in vaapi_decode_find_best_format(for example , for GBRP10, media driver will decode it with Y410.). So that can decode the clip with hwaccel at least. Otherwise it will fall back to software decoder for this kind of clip even specified using hwaccel in cmdline.

xhaihao (approved):
LGTM.

feiwan1:
submit to community:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=8445

PR #632, opened 2023-03-09
Fix tf infer failed corruption

Semmer2:
FFmpeg would corrupted when TF backend infer failed. Refine the error handle code in these case.

### DIFF
--- a/libavfilter/dnn/dnn_backend_tf.c
+++ b/libavfilter/dnn/dnn_backend_tf.c
@@ -1168,6 +1168,7 @@ int ff_dnn_execute_model_tf(const DNNModel *model, DNNExecBaseParams *exec_param
 
     ret = ff_dnn_fill_task(task, exec_params, tf_model, ctx->options.async, 1);
     if (ret != 0) {
+        av_log(ctx, AV_LOG_ERROR, "Fill task with invalid parameter(s).\n");
         av_freep(&task);
         return ret;
     }
--- a/libavfilter/dnn/dnn_backend_tf.c
+++ b/libavfilter/dnn/dnn_backend_tf.c
@@ -176,6 +176,7 @@ static int tf_start_inference(void *args)
     if (TF_GetCode(request->status) != TF_OK) {
         av_log(&tf_model->ctx, AV_LOG_ERROR, "%s", TF_Message(request->status));
         tf_free_request(infer_request);
+        av_freep(&request);
         return DNN_GENERIC_ERROR;
     }
     return 0;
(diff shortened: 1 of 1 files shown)


libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -176,9 +176,6 @@ static int tf_start_inference(void *args)
     if (TF_GetCode(request->status) != TF_OK) {
         av_log(&tf_model->ctx, AV_LOG_ERROR, "%s", TF_Message(request->status));
         tf_free_request(infer_request);
-        if (ff_safe_queue_push_back(tf_model->request_queue, request) < 0) {
-            destroy_request_item(&request);
-        }
```
Semmer2, line 181: It double calls `destroy_request_item` in line 1152, which would cause segment fault, so I delete the `destroy_request_item` here.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -453,7 +452,7 @@ static int load_tf_model(TFModel *tf_model, const char *model_filename)
         av_freep(&sess_config);
         if (TF_GetCode(tf_model->status) != TF_OK) {
             TF_DeleteGraph(tf_model->graph);
-            TF_DeleteStatus(tf_model->status);
+            tf_model->graph = NULL;
```
Semmer2, line 456: `tf_model->status` can be used in `ff_dnn_free_model_tf`, so it should not be free too early.
And the `tf_model->graph` should be set to NULL after deleting, to avoid double free in `ff_dnn_free_model_tf`.
The same reason for following change.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -480,8 +479,11 @@ static int load_tf_model(TFModel *tf_model, const char *model_filename)
         if (TF_GetCode(tf_model->status) != TF_OK)
         {
             TF_DeleteSession(tf_model->session, tf_model->status);
+            tf_model->session = NULL;
             TF_DeleteGraph(tf_model->graph);
+            tf_model->graph = NULL;
             TF_DeleteStatus(tf_model->status);
+            tf_model->status = NULL;
```
Semmer2, line 488: Since the `tf_model->session` is deleted in here, the `tf_model->status` will not be used in `ff_dnn_free_model_tf` anymore, so free it in here is totally OK.

libavfilter/dnn/dnn_backend_tf.c
```diff
@@ -940,6 +942,7 @@ DNNModel *ff_dnn_load_model_tf(const char *model_filename, DNNFunctionType func_
 
     return model;
 err:
+    model->model = tf_model;
```
Semmer2, line 948: `ff_dnn_free_model_tf` will not correctly free `tf_model`, if it's not assigned to `model->model`.

PR #635, opened 2023-03-14
Vaapi qsv dynamic frame pool

xhaihao:
A patchset to enable dynamic frame pool for QSV and VAAPI


nyanmisaka:
Been looking forward to this for a long time. This saves a huge amount of VRAM on dGPU.

4k decoding - before vs after
image

8k decoding - before vs after
image

The original fixed pool size was so large that it would quickly exhaust VRAM during multiple transcoding instances. This results in the shared memory (RAM) having to be used, which is slow as hell on dGPU.

xhaihao:
@nyanmisaka  Thanks for verifying this patchset. Note the error in pre-check wasn't caused by this patchset ( it was caused by a regression of vp8 encoding in the media driver)

xhaihao:
patchset was sent to FFmpeg mailing list https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=9639

PR #636, opened 2023-03-15
lavfi/vf_vpp_qsv: fix time_base for outlink

xhaihao:
Fixes tickets #10261, #10262

### DIFF
--- a/libavfilter/vf_vpp_qsv.c
+++ b/libavfilter/vf_vpp_qsv.c
@@ -358,7 +358,8 @@ static int config_output(AVFilterLink *outlink)
     outlink->w          = vpp->out_width;
     outlink->h          = vpp->out_height;
     outlink->frame_rate = vpp->framerate;
-    outlink->time_base  = av_inv_q(vpp->framerate);
+    outlink->time_base  = av_mul_q(inlink->time_base,
+                                   av_div_q(inlink->frame_rate, outlink->frame_rate));
 
     param.filter_frame  = NULL;
     param.num_ext_buf   = 0;


dvrogozh (approved):
This PR works for us. We verified that both reported issues are gone after it. I suggest to proceed submitting patch to ffmpeg mailing list.

xhaihao:
@dvrogozh  I worked out another patch (https://github.com/intel-media-ci/ffmpeg/pull/637) and will submit #637 to FFmpeg ML. You may use #636 or #637 in your  docker.

dvrogozh:
> @dvrogozh I worked out another patch (https://github.com/intel-media-ci/ffmpeg/pull/637) and will submit https://github.com/intel-media-ci/ffmpeg/pull/637 to FFmpeg ML. You may use https://github.com/intel-media-ci/ffmpeg/pull/636 or https://github.com/intel-media-ci/ffmpeg/pull/637 in your docker

Which patch will actually be submitted to mailing list? We eventually did not verify 637 if it will go instead of 636 then we need to redo the test tomorrow.

xhaihao:
> > @dvrogozh I worked out another patch (#637) and will submit #637 to FFmpeg ML. You may use #636 or #637 in your docker
> 
> Which patch will actually be submitted to mailing list? We eventually did not verify 637 if it will go instead of 636 then we need to redo the test tomorrow.

I will submit ***637*** to ML.  This is because the timestamp for EOF still might be wrong even if applying 636.

dvrogozh:
Will you submit 636 and 637 or only 637? these patches don't seem to be mutually exclusive.

xhaihao:
> Will you submit 636 and 637 or only 637? these patches don't seem to be mutually exclusive.

637 only.  We may set time_base to any valid value, hence 636 is not needed after applying 637.

dvrogozh:
> We may set time_base to any valid value, hence 636 is not needed after applying 637.

I am confused. 636 clearly shows that we can't set it to any value since arbitrary values cause issues. I honestly don't know what time_base is supposed to do, but it should have some very specific meaning, not being arbitrary.

xhaihao:
> > We may set time_base to any valid value, hence 636 is not needed after applying 637.
> 
> I am confused. 636 clearly shows that we can't set it to any value since arbitrary values cause issues. I honestly don't know what time_base is supposed to do, but it should have some very specific meaning, not being arbitrary.

Yes, we may set different time_base (num, den), e.g. for a 60 fps video, we may use (1, 60), (1, 120), (1, 240) ... as time_base., we should rescale the timestamp based on the given time_base. The root cause for your issues is that a timestamp is set with wrong time_base.

pts in the following code is gotten from inlink,  it is in time_base units on the inlink, but we set this pts to outlink directly.

eof:
    ff_outlink_set_status(outlink, status, pts);                                                                                                                                               
    return 0;

So we should rescale pts for outlink before setting this pts to outlink 

pts = av_rescale_q(pts, inlink->time_base, outlink->time_base);

So even if 636 is applied, 637 is still needed when the time_base on outlink is different from the one on inlink.  On the other hand, when 637 is applied, 636 is not needed anymore because the timestamp is now in time_base units on outlink.

PR #638, opened 2023-03-20
liavfi/dnn: Add OpenVINO C API 2.0 support

Semmer2:
OpenVINO was released in March 2022, which introduced many new feature. Will implement current OpenVINO features with new 2.0 APIs, and adding new features with new APIs.


Semmer2:
This is a draft.

PR #643, opened 2023-04-12
lavfi/overlay_vaapi: set input region for overlay video to NULL

xhaihao:
Otherwise the region is copied from main video. Note NULL imples the whole overlay video will be processed.


xhaihao:
See https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230417071603.226543-1-haihao.xiang@intel.com/

PR #649, opened 2023-05-06
lavc/qsv: try another mfx implementation for internal session on Windows

xhaihao:
There isn't a requirement on implemenation for internal session, so we may try another implementation if the preferred one is not supported


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230515060403.68363-1-haihao.xiang@intel.com/

PR #663, opened 2023-06-28
libavfilter/vf_dnn_detect: bbox index may bigger than bbox number

wenbinc-Bin:
Fix a bug that queried bbox index may bigger than bbox's total number.


wenbinc-Bin:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230717053301.2389494-1-wenbin.chen@intel.com/

PR #671, opened 2023-08-09
lavfi/dnn: Add OpenVINO API 2.0 support

wenbinc-Bin:
OpenVINO API 2.0 was released in March 2022, which introduced new features.
This commit implements current OpenVINO features with new 2.0 APIs. And will add other features in API 2.0.
Please add installation path, which include openvino.pc, to PKG_CONFIG_PATH mannually for new OpenVINO libs config.

### DIFF
--- a/libavfilter/dnn/dnn_backend_openvino.c
+++ b/libavfilter/dnn/dnn_backend_openvino.c
@@ -32,7 +32,11 @@
 #include "libavutil/detection_bbox.h"
 #include "../internal.h"
 #include "safe_queue.h"
+#if HAVE_OPENVINO2
+#include <openvino/c/openvino.h>
+#else
 #include <c_api/ie_c_api.h>
+#endif
 #include "dnn_backend_common.h"
 
 typedef struct OVOptions{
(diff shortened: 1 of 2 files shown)


libavfilter/dnn/dnn_backend_openvino.c
```diff
@@ -51,9 +55,20 @@ typedef struct OVContext {
 typedef struct OVModel{
     OVContext ctx;
     DNNModel *model;
+#if HAVE_OPENVINO2
```
guoyejun, line 58: looks that we support both ov2 and ov1?

wenbinc-Bin: We support openvino. If it has openvino2 API, then we use openvino2 API.

xhaihao: Will it work if there is only ov1 in the runtime but user builds FFmpeg against ov2 or inverse ?

wenbinc-Bin: No it won't work. We need to make sure it uses same env when both compiling and running.

wenbinc-Bin:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230815082631.2648570-1-wenbin.chen@intel.com/

PR #672, opened 2023-08-10
lavc/qsvenc_hevc: restore the default gop size

xhaihao:
commit a3c0a3e changed the default settings and expected the runtime can choose a best value. However the runtime doesn't set a valid gop size for hevc encoder, hence the output steam is non-seekable, which is inconvenient to user [1][2]

[1] https://github.com/intel/media-driver/issues/1576
[2] https://ffmpeg.org/pipermail/ffmpeg-user/2023-August/056716.html


wenbinc-Bin:
LGTM

xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20230811063541.536421-1-haihao.xiang@intel.com/

PR #676, opened 2023-08-28
avcodec/jpegxl_parser: fix a compile error

tong1wu:
Compiler: MSVC 14.35.32215
Error type: error C2099: initializer is not a constant
Related commit: 0c0dd23 avcodec/jpegxl_parser: add JPEG XL parser


tong1wu:
Merged into upstream. Closed.

PR #680, opened 2023-09-15
configure: don't warn deprecated symbols from libvpl

xhaihao:
Some symbols are deprecated in libvpl (e.g. MFX_EXTBUFF_VPP_DENOISE2 is used to replace MFX_EXTBUFF_VPP_DENOISE, but MFX_EXTBUFF_VPP_DENOISE2 is not supported in MediaSDK runtime). In order to support MediaSDK runtime, we have to use these deprecated symbols, so add -DMFX_DEPRECATED_OFF to CFLAGS to silence the corresponding compilation warnings


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20231127081152.2437803-1-haihao.xiang@intel.com/

PR #683, opened 2023-10-11
lavfi/vf_vpp_qsv: set the default value of async_depth option to 4

xhaihao:
Both qsv encoders and decoders use 4 as default value of async_depth, let's use 4 as default value for vpp_qsv filter too.


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=10131

PR #690, opened 2023-11-21
lavc/vaapi_encode_av1: Add qp option explicitly to set base q index

feiwan1:
Keep same way with librav1e/libsvtav1/qsv_av1.. to make it more acceptable instead of using global option "-global_quality".

Fix #10615


xhaihao (approved):
LGTM

feiwan1:
submit to https://patchwork.ffmpeg.org/project/ffmpeg/patch/20231127005801.3440746-1-fei.w.wang@intel.com/

PR #700, opened 2024-01-03
Openvino

wenbinc-Bin:
1. Add automatic ports detection
2. change width, height and channel to dims[4]


uartie:
please fix conflicts or close. thanks.

PR #701, opened 2024-01-17
lavc/qsvenc: Add workaround for VP9 keyframe

xhaihao:
The runtime doesn't set keyframe flag on a returned keyframe, which makes the coded stream unseekable.


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20240126042822.111719-1-haihao.xiang@intel.com/

PR #703, opened 2024-01-22
lavc/vaapi_encode: Enable Macroblock based bitrate control

feiwan1:


libavcodec/vaapi_encode.c
```diff
@@ -2052,6 +2060,7 @@ static av_cold int vaapi_encode_init_rate_control(AVCodecContext *avctx)
 #if VA_CHECK_VERSION(1, 1, 0)
             .ICQ_quality_factor = av_clip(rc_quality, 1, 51),
             .max_qp             = (avctx->qmax > 0 ? avctx->qmax : 0),
+            .rc_flags.bits.mb_rate_control = ctx->mbbrc ? 1 : 2,
```
xhaihao, line 2063: Why do we not use the default value (0) ?

feiwan1: MBBRC brings higher quality to objects in picture, but drops frame level PSNR. So I'd prefer to enable/disable it in ffmpeg if user really understand its pros and cons instread of decided automatically by driver. And for VPL, it also disabled by default for some hw/codecs.

xhaihao: > And for VPL, it also disabled by default for some hw/codecs.

So it is enabled for some hw/codecs, right ?

feiwan1: > > And for VPL, it also disabled by default for some hw/codecs.
> 
> So it is enabled for some hw/codecs, right ?

Only auto(Unknow) and disable(OFF) by default from ffmpeg vpl log.

xhaihao: Does setting mb_rate_control to 1 or 2 make sense while the underlying driver doesn't support VA_RC_MB ?

feiwan1: There is code to check supported VA_RC_MB, if not, mb_rate_control will be set to 2(disable) only.

xhaihao (approved):
LGTM

feiwan1:
submit to https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=10708.

PR #704, opened 2024-02-01
qsvenc: update the selection of bitrate control method

xhaihao:
The default method is changed to CQP


xhaihao:
https://ffmpeg.org/pipermail/ffmpeg-devel/2024-February/321088.html

PR #713, opened 2024-02-29
lavu/hwcontext_qsv: Join the download/upload session to the main session

xhaihao:
This may reduce the number of internal threads when using hwupload or hwdownload filter.


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=11085

PR #716, opened 2024-03-05
lavu/vulkan_filter: fix input format

xhaihao:
Otherwise s->input_format is always yuv420p, which will result in wrong output format in vulkan filters.

This fixes invalid output format for hwframe download in the command below:
./ffmpeg -init_hw_device vulkan -f lavfi \
	-i testsrc=duration=1,format=nv12 \
	-vf 'hwupload,format=vulkan,scale_vulkan=1024:768,hwdownload,format=nv12' \
	-f null -


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20240306052559.2762518-1-haihao.xiang@intel.com/

PR #721, opened 2024-03-10
lavc/qsvdec: Do not print warning for a NULL bitstream pointer

xhaihao:
It is the expected behavior to return MFX_ERR_MORE_DATA when calling MFXVideoDECODE_DecodeFrameAsync with a NULL bitstream pointer.


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=11084

PR #722, opened 2024-03-11
libavcodec/qsvdec: add dump_extra bsf to h264/hevc_qsv decoder

wenbinc-Bin:
According to VPL doc:

"Before decoding the first frame, a sequence header must be present. The function skips any bitstreams before it encounters the new sequence header."

When decode a live video, sequence header is passed through sdp and video stream may not have sequence header. In this case qsv_decoder cannot decode header and keep waiting for more data.

For example:
push stream:
ffmpeg -v verbose -re -stream_loop -1 -i input.mp4 -c:v hevc_qsv \
 -bsf:v extract_extradata=remove=1 -f rtsp \
 rtsp://localhost:8554/mystream

receive stream:
ffmpeg -v verbose -hwaccel qsv -i rtsp://localhost:8554/mystream \
 -f null -

Header data is stored in extradata, so add "dump_extra" bsf into qsvdec, and then qsv decoder works on video stream without sequence header.

### DIFF
--- a/libavcodec/qsvdec.c
+++ b/libavcodec/qsvdec.c
@@ -1160,7 +1160,7 @@ static const AVOption hevc_options[] = {
         { "off",     NULL, 0, AV_OPT_TYPE_CONST, { .i64 = MFX_GPUCOPY_OFF },     0, 0, VD, .unit = "gpu_copy"},
     { NULL },
 };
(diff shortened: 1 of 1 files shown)


xhaihao (approved):
LGTM

PR #725, opened 2024-03-15
Add pad_vaapi & drawbox_vaapi filters

xhaihao:


wenbinc-Bin:
Do we need to draw outer_frame every time? If we draw outer_frame only once and reuse it, will it be better?

xhaihao:
> Do we need to draw outer_frame every time? If we draw outer_frame only once and reuse it, will it be better?

Thanks for the comment, I fixed it in the new version.

wenbinc-Bin:
LGTM. Thanks

xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=11175

PR #731, opened 2024-03-28
avcodec/hevc_ps: fix a potential memcmp mismatch

tong1wu:
HEVCHdrParams* receives a pointer which points to a dynamically allocated memory block. It causes the memcmp always returning 1. Add a function to do the comparision. A condition is also added to avoid malloc(0).


xhaihao:
the patch was merged, see http://git.videolan.org/?p=ffmpeg.git;a=commit;h=6bf17136a2bc1e6f52ea9cd27a6dcaab648efe5c

PR #739, opened 2024-04-22
lavc/vp9: Fix regression introduced in 0ba05857

xhaihao:
ff_progress_frame_await() is called when hardware acceleration method is used, however ff_progress_frame_report() isn't called. A decoding thread for VP9 might get stuck.


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20240423043258.30883-1-haihao.xiang@intel.com/

PR #740, opened 2024-04-23
lavfi/qsv: Copy metadata fields from the given input 

xhaihao:
Currently it always copies the metadata fields from the last input when
there are multiple inputs for the filter. For example, the medadata
fields from input1 is copied to the output for overlay_qsv filter,
however for regular overlay filters, the medata fields from input0 are
copied to the output. With this fix, we may copy the medadata fields
from input0 to the ouput as well.

### DIFF
--- a/libavfilter/qsvvpp.c
+++ b/libavfilter/qsvvpp.c
@@ -441,11 +441,6 @@ static QSVFrame *submit_frame(QSVVPPContext *s, AVFilterLink *inlink, AVFrame *p
                 av_frame_free(&qsv_frame->frame);
                 return NULL;
             }
-
-            if (av_frame_copy_props(qsv_frame->frame, picref) < 0) {
-                av_frame_free(&qsv_frame->frame);
-                return NULL;
-            }
         } else
             qsv_frame->frame = av_frame_clone(picref);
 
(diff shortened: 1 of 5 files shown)


libavfilter/qsvvpp.c
```diff
@@ -981,7 +964,7 @@ int ff_qsvvpp_filter_frame(QSVVPPContext *s, AVFilterLink *inlink, AVFrame *picr
         aframe.frame->frame = NULL;
     };
 
-    if (!picref)
+    if (!picref || !propref)
```
wenbinc-Bin: Is there a possibility that the first frame we received is not from input 0, so that propref is NULL but the picref is not NULL. The function return and the first frame is lost?

wenbinc-Bin: How about moving propref check to just before calling av_frame_copy_props()?

xhaihao: Updated, thanks for your comments.

wenbinc-Bin:
LGTM

xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20240425051809.283325-1-haihao.xiang@intel.com/

PR #741, opened 2024-04-25
lavc/qsvdec: Allow decoders to export crop information

feiwan1:
Correct frame resolution when it change inside a sequence which has max width/height in its syntax. In this case, it will not trigger reinit process, and use the 1st frame's resolution for all frames.

### DIFF
--- a/libavcodec/qsvdec.c
+++ b/libavcodec/qsvdec.c
@@ -856,6 +856,10 @@ static int qsv_decode(AVCodecContext *avctx, QSVContext *q,
             else
                 frame->flags &= ~AV_FRAME_FLAG_KEY;
         }
+        frame->crop_left   = outsurf->Info.CropX;
+        frame->crop_top    = outsurf->Info.CropY;
+        frame->crop_right  = outsurf->Info.Width - (outsurf->Info.CropX + outsurf->Info.CropW);
+        frame->crop_bottom = outsurf->Info.Height - (outsurf->Info.CropY + outsurf->Info.CropH);
 
         /* update the surface properties */
         if (avctx->pix_fmt == AV_PIX_FMT_QSV)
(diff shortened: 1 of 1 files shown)


libavcodec/qsvdec.c
```diff
@@ -851,6 +851,10 @@ static int qsv_decode(AVCodecContext *avctx, QSVContext *q,
             else
                 frame->flags &= ~AV_FRAME_FLAG_KEY;
         }
+        if (frame->width != outsurf->Info.CropW || frame->height != outsurf->Info.CropH) {
+            avctx->width  = frame->width  = outsurf->Info.CropW;
+            avctx->height = frame->height = outsurf->Info.CropH;
+        }
```
xhaihao: widthxheight is the coded dimensions (in pixel) which is different from the cropping rectangle, see https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/frame.h#L437-L446

feiwan1: OK, export crop props seems a more generic way, which will also take care of real crop case. I will try to modify it.

libavutil/hwcontext_qsv.c
```diff
@@ -1317,8 +1317,6 @@ static int qsv_get_buffer(AVHWFramesContext *ctx, AVFrame *frame)
 
     frame->data[3] = frame->buf[0]->data;
     frame->format  = AV_PIX_FMT_QSV;
-    frame->width   = ctx->width;
-    frame->height  = ctx->height;
```
xhaihao: qsv_get_buffer is used in qsv based filters too, does this change impact these filters ?

xhaihao (approved):
LGTM

feiwan1:
https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=11837

PR #749, opened 2024-06-04
lavu/hwcontext_vulkan: Support write on drm frame

xhaihao:
Otherwise nothing is written into the destination when a write mapping is requested.

For example, a vulkan frame mapped from a drm frame (which is wrapped as a vaapi frame in the example) is used as the output of scale_vulkan filter, it always gets a green screen without this patch.

ffmpeg -init_hw_device vaapi=va -init_hw_device vulkan=vulkan@va -filter_hw_device vulkan -f lavfi -i testsrc=size=352x288,format=nv12 -vf
"hwupload,scale_vulkan,hwmap=derive_device=vaapi:reverse=1,format=vaapi,hwdownload,format=nv12" -f nut - | ffplay -


xhaihao:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20240611045547.334627-1-haihao.xiang@intel.com/

PR #751, opened 2024-06-12
libavfilter/vf_python: Add video python filter

wenbinc-Bin:
Now many image algorithms are implemented in python. Add a filter that can call python script to process video frames. User need to pass script name, filter function and init function ( optional) to python filter. AVFrame will be transferred to a Tuple of numpy.ndarray and be passed to python script.

for example we have python script
/****ffmpeg-python.py****/
offset = 0
def init(init_str=None):
  global offset
  offset = int(init_str)

def filter_func(frame):
  for plane in frame:
    plane += offset
  return frame
/******end file************/

Add the path of script file into environment variable PYTHONPATH. we can call ffmpeg:
ffmpeg -i input.264 -vf format=rgb24, \
  video_python=module=ffmpeg-python:init_func=init:init_param="10": \
  filter_func=filter_func output.mp4

Which adds offset to each pixel of the frame.

### DIFF
--- a/configure
+++ b/configure
@@ -335,6 +335,7 @@ External library support:
   --enable-vapoursynth     enable VapourSynth demuxer [no]
   --disable-xlib           disable xlib [autodetect]
   --disable-zlib           disable zlib [autodetect]
+  --enable-python          enable Python code [no]
 
   The following libraries provide various hardware acceleration features:
   --disable-amf            disable AMF video encoding code [autodetect]
@@ -1980,6 +1981,7 @@ EXTERNAL_LIBRARY_LIST="
     openssl
     pocketsphinx
     vapoursynth
+    python
 "
 
 HWACCEL_AUTODETECT_LIBRARY_LIST="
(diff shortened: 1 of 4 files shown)


xhaihao:
s/algrithms/algorithms and s/transfered/transferred in the commit log

xhaihao:
./configure --enable-python worked but make failed
libavfilter/vf_python.c:54:10: fatal error: numpy/arrayobject.h: No such file or directory
   54 | #include "numpy/arrayobject.h"
      |          ^~~~~~~~~~~~~~~~~~~~~

wenbinc-Bin:
> ./configure --enable-python worked but make failed libavfilter/vf_python.c:54:10: fatal error: numpy/arrayobject.h: No such file or directory 54 | #include "numpy/arrayobject.h" | ^~~~~~~~~~~~~~~~~~~~~

you need to "pip install numpy"

wenbinc-Bin:
> ./configure --enable-python worked but make failed libavfilter/vf_python.c:54:10: fatal error: numpy/arrayobject.h: No such file or directory 54 | #include "numpy/arrayobject.h" | ^~~~~~~~~~~~~~~~~~~~~

I add to commit log and comment.

PR #764, opened 2024-08-26
cbs_av1: Copy loop filter and segment parameters of repeat frame fro…

feiwan1:
…m its mapped frame

Fixes #11151


feiwan1:
https://patchwork.ffmpeg.org/project/ffmpeg/patch/20240828120144.549180-1-fei.w.wang@intel.com/

